PR #1, opened 2019-05-16
Add HealthComponent

e-aakash:
For now, just moved the HealthComponent of Logic/Health to the new module.

### DIFF
--- a/module.txt
+++ b/module.txt
@@ -4,6 +4,6 @@
     "author" : "",
     "displayName" : "Health",
     "description" : "This module provides basic implementation of Health",
-    "dependencies" : [],
+    "dependencies" : [{"id": "ModuleTestingEnvironment","minVersion": "0.1.0", "optional":true}],
     "serverSideOnly" : false
 }
--- a/src/main/java/org/terasology/HealthComponent.java
+++ b/src/main/java/org/terasology/HealthComponent.java
@@ -19,29 +19,46 @@ import org.terasology.entitySystem.Component;
 import org.terasology.network.Replicate;
 import org.terasology.rendering.nui.properties.TextField;
 
+/**
+ * Provides Health to entity attached with HealthComponent. Contains the parameters
+ * required for all health related events.
+ */
 public class HealthComponent implements Component {
 
+    /** Maximum allowed health, capped to this if exceeding this value. */
     @Replicate
     public int maxHealth = 20;
+
+    /** Amount of health restored in each regeneration tick. */
     @Replicate
     public float regenRate;
+
+    /** Time delay before regeneration starts. */
     @Replicate
     public float waitBeforeRegen;
 
+    /** Falling speed threshold above which damage is inflicted to entity. */
     @Replicate
     public float fallingDamageSpeedThreshold = 20;
+
+    /** Horizontal speed threshold above which damage is inflicted to entity. */
     @Replicate
     public float horizontalDamageSpeedThreshold = 20;
+
+    /** The multiplier used to calculate damage when horizontal or vertical threshold is crossed. */
     @Replicate
     public float excessSpeedDamageMultiplier = 10f;
 
+
+    /** The current value of health. */
     @Replicate
     @TextField
     public int currentHealth = 20;
 
-    // Regen info
+    /** Next tick time that will trigger regeneration. */
     public long nextRegenTick;
 
+    /** Used to send Destroy event when health breaches zero. */
     public boolean destroyEntityOnNoHealth;
 
 }
--- a/src/main/java/org/terasology/HealthComponent.java
+++ b/src/main/java/org/terasology/HealthComponent.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2019 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology;
+
+import org.terasology.entitySystem.Component;
+import org.terasology.network.Replicate;
+import org.terasology.rendering.nui.properties.TextField;
+
+public class HealthComponent implements Component {
+
+    @Replicate
+    public int maxHealth = 20;
+    @Replicate
+    public float regenRate;
+    @Replicate
+    public float waitBeforeRegen;
+
+    @Replicate
+    public float fallingDamageSpeedThreshold = 20;
+    @Replicate
+    public float horizontalDamageSpeedThreshold = 20;
+    @Replicate
+    public float excessSpeedDamageMultiplier = 10f;
+
+    @Replicate
+    @TextField
+    public int currentHealth = 20;
+
+    // Regen info
+    public long nextRegenTick;
+
+    public boolean destroyEntityOnNoHealth;
+
+    public HealthComponent() {
+    }
+
+    public HealthComponent(int maxHealth, float regenRate, float waitBeforeRegen) {
+        this.maxHealth = maxHealth;
+        this.currentHealth = maxHealth;
+        this.regenRate = regenRate;
+        this.waitBeforeRegen = waitBeforeRegen;
+    }
+
+}
--- a/src/main/java/org/terasology/HealthComponent.java
+++ b/src/main/java/org/terasology/HealthComponent.java
@@ -44,14 +44,4 @@ public class HealthComponent implements Component {
 
     public boolean destroyEntityOnNoHealth;
 
-    public HealthComponent() {
-    }
-
-    public HealthComponent(int maxHealth, float regenRate, float waitBeforeRegen) {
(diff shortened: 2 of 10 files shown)


src/main/java/org/terasology/HealthComponent.java
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology;
+
+import org.terasology.entitySystem.Component;
+import org.terasology.network.Replicate;
+import org.terasology.rendering.nui.properties.TextField;
+
+public class HealthComponent implements Component {
```
Cervator: Well-written class level javadoc can be useful in very central components like this. Think long term: what might somebody like to see when they're fishing around for relevant components and run into Health?

src/main/java/org/terasology/HealthComponent.java
```diff
@@ -0,0 +1,57 @@
...
+    @Replicate
+    public float horizontalDamageSpeedThreshold = 20;
+    @Replicate
+    public float excessSpeedDamageMultiplier = 10f;
+
+    @Replicate
+    @TextField
+    public int currentHealth = 20;
+
+    // Regen info
```
Cervator: Prefer one-liner javadoc comments IMHO and document each field you think can be meaningfully enhanced. Example:

`/** Next tick time that will trigger regeneration */

src/main/java/org/terasology/HealthComponent.java
```diff
@@ -0,0 +1,57 @@
...
+    public long nextRegenTick;
+
+    public boolean destroyEntityOnNoHealth;
+
+    public HealthComponent() {
```
Cervator: As a practice it may be wise to leave a comment on the no-arg constructor to highlight the reason it is here: if it isn't and there is an arg-constructor then the "invisible" no-arg constructor disappears. That then breaks stuff as some of the magic relies on there always being a no-arg constructor :-)

src/main/java/org/terasology/HealthComponent.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public HealthComponent(int maxHealth, float regenRate, float waitBeforeRegen) {
```
Cervator: I figure this constructor is here because it already existed - I wonder how frequently it is used and if it is really justified for it to exist? It might be a good practice for us to always encourage the pattern of setting whichever exact fields you need rather than use a custom constructor

```java
HealthComponent health;
health.maxHealth = 123;
health.regenRate = 42;
//apply the component to something
```

Cervator (changes requested):
Throwing in some general comments hoping bits of it might be useful :-)

Another suggestion: Make a paired draft PR for the engine that removes anything this PR adds to the Health module, then update it over time. No intent to merge the engine PR for some time, but it might be good for visibility and being prepared. Then finally when Health is ready switch the engine PR out of draft mode (and/or recreate it fresh depending on how stale that branch went)

src/main/java/org/terasology/HealthComponent.java
e-aakash: After searching in the mega workspace, I couldn't find any usage of this constructor. I am hence removing this constructor.
And, since the unique constructor is removed, the default constructor is not needed explicitly.

Cervator (approved):

src/test/java/org/terasology/logic/health/HealEventTest.java
```diff
@@ -0,0 +1,70 @@
+/*
```
iaronaraujo, line 1: I am not sure if this test should be here or in the MTE module.

iaronaraujo (approved):

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -0,0 +1,155 @@
...
+@RegisterSystem(RegisterMode.AUTHORITY)
+public class HealthAuthoritySystem extends BaseComponentSystem implements UpdateSubscriberSystem {
+
+    private static final Logger logger = LoggerFactory.getLogger(HealthAuthoritySystem.class);
+
+    @In
+    private EntityManager entityManager;
+
+    @In
+    private org.terasology.engine.Time time;
```
Cervator, line 59: Any reason for having this fully qualified?

module.txt
```diff
@@ -4,6 +4,6 @@
     "author" : "",
```
Cervator, line 4: Feel free to toss an author name in there :-)

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+        TFloatIterator multiplierIter = multipliers.iterator();
+        while (multiplierIter.hasNext()) {
+            total *= multiplierIter.next();
+        }
+        return TeraMath.floorToInt(total);
+
+    }
+
+    /** Handles DoHeal event to increase health of entity */
```
Cervator, line 143: One-liner javadoc is good for simple fields, but methods should always use the expanded version including doc for parameters etc :-)

src/main/java/org/terasology/logic/health/event/DoHealEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.logic.health.event;
+
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.Event;
+
+/**
+ * Send this event to an entity to heal it.
```
Cervator, line 22: "to heal it ..." - fully? Or could a negative number be hurtful instead? Maybe that's covered in the System

src/main/java/org/terasology/logic/health/event/FullHealthEvent.java
```diff
@@ -0,0 +1,36 @@
...
+package org.terasology.logic.health.event;
+
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.Event;
+
+/**
+ * Event sent upon an entity reaching full health if previously on less than full health.
+ *
+ */
+public class FullHealthEvent implements Event {
```
Cervator, line 25: Might this be more descriptive as `OnFullHealthEvent` ?

Would also put it more in line with `OnHealedEvent` - although then I'd wonder if this would work better as `OnFullyHealedEvent` plus wonder what the implications of both triggering would be

src/main/java/org/terasology/logic/health/event/HealthChangedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+        this.instigator = instigator;
+        this.change = change;
+    }
+
+    public EntityRef getInstigator() {
+        return instigator;
+    }
+
+    /**
+     * @return The amount by which health changed. This is capped, so if the entity received 9999 damage and only had 10 health, it will be -10.
```
Cervator, line 40: Is it though? This method just seems to blindly return the change - I don't see any capping here? It could happen elsewhere but then that'd make this javadoc somewhat presumptuous :-)

It also helps to still write out both statements separately - the method description and the return value. Sometimes that can be tricky to write in a meaningful way split into two fields but since they each go to a different place in displayed javadoc it is ideal to have both (IMHO anyway)

An alternative might be to throw a boolean into a single event that indicates whether or not it resulted in a full (or partial) heal

src/test/java/org/terasology/logic/health/HealEventTest.java
```diff
@@ -0,0 +1,132 @@
...
+        player.addComponent(new PlayerCharacterComponent());
+        player.addComponent(healthComponent);
+
+        TestEventReceiver<BeforeHealEvent> receiver = new TestEventReceiver<>(getHostContext(), BeforeHealEvent.class,
+                (event, entity) -> {
+                    event.add(10);
+                    event.subtract(5);
+                    event.multiply(2);
+                });
+        // expected = ( initial:10 + 10 - 5 ) * 2 = 30
```
Cervator, line 127: I'd rewrite this ever so slightly for clarity. Like:

```suggestion
        // Expected value is ( initial:10 + 10 - 5 ) * 2 == 30
```

Cervator (approved):
Left some more comments but mostly trivial / future thoughts so not going to block merging from it. Will push to merge without a merge commit so you can continue in your next branch without any updates needed. Can address any actionable review comments in a new commit during that effort :-)

Nice tests! They actually work for me for some reason, without hacking my workspace. I don't understand, but ... ¯\_(ツ)_/¯

Haven't tested in-game as I don't know if that's ready yet.

src/main/java/org/terasology/logic/health/event/HealthChangedEvent.java
e-aakash: I think it does show the changed amount because the capping is done not in the event but in the method calling the event.
https://github.com/Terasology/Health/pull/1/files#diff-e0c50392da47b79c5d45c9c66a5ec5feR112

So, i think the javadoc should be modified to say so.

src/main/java/org/terasology/logic/health/event/DoHealEvent.java
e-aakash: A negative value is supposed to cause damage, but not working at present. Will update the Javadoc when the issue is solved.

src/main/java/org/terasology/logic/health/event/FullHealthEvent.java
e-aakash: Renaming it as `OnFullyHealedEvent`, seems more apt for full health event.

PR #2, opened 2019-05-22
Add damage events

e-aakash:
Add damage events and update HealthAuthoritySystem.

### DIFF
--- a/src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
@@ -141,10 +141,11 @@ public class HealthAuthoritySystem extends BaseComponentSystem implements Update
     private void doHeal(EntityRef entity, int healAmount, EntityRef instigator) {
         HealthComponent health = entity.getComponent(HealthComponent.class);
         if (health != null) {
-            int healedAmount = Math.min(health.currentHealth + healAmount, health.maxHealth) - health.currentHealth;
-            health.currentHealth += healedAmount;
+            int cappedHealth = Math.min(health.currentHealth + healAmount, health.maxHealth);
+            int cappedHealAmount = cappedHealth - health.currentHealth;
+            health.currentHealth = cappedHealth;
             entity.saveComponent(health);
-            entity.send(new OnHealedEvent(healAmount, healedAmount, instigator));
+            entity.send(new OnHealedEvent(healAmount, cappedHealAmount, instigator));
             if (health.currentHealth == health.maxHealth) {
                 entity.send(new OnFullyHealedEvent(instigator));
             }
@@ -205,10 +206,6 @@ public class HealthAuthoritySystem extends BaseComponentSystem implements Update
      */
     @ReceiveEvent(components = HealthComponent.class, netFilter = RegisterMode.AUTHORITY)
     public void onAttackEntity(AttackEvent event, EntityRef targetEntity) {
-        damageEntity(event, targetEntity);
-    }
-
-    static void damageEntity(AttackEvent event, EntityRef targetEntity) {
         int damage = 1;
         Prefab damageType = EngineDamageTypes.PHYSICAL.get();
         // Calculate damage from item
@@ -233,11 +230,11 @@ public class HealthAuthoritySystem extends BaseComponentSystem implements Update
             ghost = (characterMovementComponent.mode == MovementMode.GHOSTING);
         }
         if ((health != null) && !ghost) {
-            int damagedAmount = health.currentHealth - Math.max(health.currentHealth - damageAmount, 0);
-            health.currentHealth -= damagedAmount;
+            int cappedDamage = Math.min(health.currentHealth, damageAmount);
+            health.currentHealth -= cappedDamage;
             health.nextRegenTick = time.getGameTimeInMs() + TeraMath.floorToInt(health.waitBeforeRegen * 1000);
             entity.saveComponent(health);
-            entity.send(new OnDamagedEvent(damageAmount, damagedAmount, damageType, instigator));
+            entity.send(new OnDamagedEvent(damageAmount, cappedDamage, damageType, instigator));
             if (health.currentHealth == 0 && health.destroyEntityOnNoHealth) {
                 entity.send(new DestroyEvent(instigator, directCause, damageType));
             }
@@ -275,7 +272,7 @@ public class HealthAuthoritySystem extends BaseComponentSystem implements Update
      * @param characterSounds Component having sound settings.
      */
     @ReceiveEvent
-    public void onDamaged(OnDamagedEvent event, EntityRef entity, CharacterSoundComponent characterSounds) {
+    public void onDamaged(OnDamagedEvent event, EntityRef entity, CharacterSoundComponent characterSounds)  {
         if (characterSounds.lastSoundTime + CharacterSoundSystem.MIN_TIME < time.getGameTimeInMs()) {
 
             // play the sound of damage hitting the character for everyone
@@ -307,17 +304,11 @@ public class HealthAuthoritySystem extends BaseComponentSystem implements Update
      * @param event VerticalCollisionEvent sent when falling speed threshold is crossed.
      * @param entity The entity which is damaged due to falling.
      */
-    @ReceiveEvent(components = {HealthComponent.class})
-    public void onLand(VerticalCollisionEvent event, EntityRef entity) {
-        HealthComponent health = entity.getComponent(HealthComponent.class);
+    @ReceiveEvent
+    public void onLand(VerticalCollisionEvent event, EntityRef entity, HealthComponent health) {
         float speed = Math.abs(event.getVelocity().y);
 
-        if (speed > health.fallingDamageSpeedThreshold) {
-            int damage = (int) ((speed - health.fallingDamageSpeedThreshold) * health.excessSpeedDamageMultiplier);
-            if (damage > 0) {
-                checkDamage(entity, damage, EngineDamageTypes.PHYSICAL.get(), EntityRef.NULL, EntityRef.NULL);
-            }
-        }
+        highSpeedDamage(speed, entity, health.fallingDamageSpeedThreshold, health.excessSpeedDamageMultiplier);
     }
 
     /**
@@ -326,16 +317,18 @@ public class HealthAuthoritySystem extends BaseComponentSystem implements Update
      * @param event HorizontalCollisionEvent sent when "falling horizontally".
      * @param entity Entity which is damaged on "horizontal fall".
      */
-    @ReceiveEvent(components = {HealthComponent.class})
-    public void onCrash(HorizontalCollisionEvent event, EntityRef entity) {
-        HealthComponent health = entity.getComponent(HealthComponent.class);
-
+    @ReceiveEvent
+    public void onCrash(HorizontalCollisionEvent event, EntityRef entity, HealthComponent health) {
         Vector3f vel = new Vector3f(event.getVelocity());
         vel.y = 0;
         float speed = vel.length();
 
-        if (speed > health.horizontalDamageSpeedThreshold) {
-            int damage = (int) ((speed - health.horizontalDamageSpeedThreshold) * health.excessSpeedDamageMultiplier);
+        highSpeedDamage(speed, entity, health.horizontalDamageSpeedThreshold, health.excessSpeedDamageMultiplier);
+    }
+
+    private void highSpeedDamage(float speed, EntityRef entity, float threshold, float damageMultiplier) {
+        if (speed > threshold) {
+            int damage = (int) ((speed - threshold) * damageMultiplier);
             if (damage > 0) {
                 checkDamage(entity, damage, EngineDamageTypes.PHYSICAL.get(), EntityRef.NULL, EntityRef.NULL);
             }
--- a/src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
@@ -21,20 +21,32 @@ import gnu.trove.list.TFloatList;
 import gnu.trove.list.TIntList;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.terasology.audio.StaticSound;
+import org.terasology.audio.events.PlaySoundEvent;
+import org.terasology.audio.events.PlaySoundForOwnerEvent;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.entitySystem.event.ReceiveEvent;
+import org.terasology.entitySystem.prefab.Prefab;
 import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterMode;
 import org.terasology.entitySystem.systems.RegisterSystem;
 import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
-import org.terasology.logic.health.event.BeforeHealEvent;
-import org.terasology.logic.health.event.DoHealEvent;
-import org.terasology.logic.health.event.FullHealthEvent;
-import org.terasology.logic.health.event.OnHealedEvent;
+import org.terasology.logic.characters.CharacterMovementComponent;
+import org.terasology.logic.characters.CharacterSoundComponent;
+import org.terasology.logic.characters.CharacterSoundSystem;
+import org.terasology.logic.characters.MovementMode;
+import org.terasology.logic.characters.events.AttackEvent;
+import org.terasology.logic.characters.events.HorizontalCollisionEvent;
+import org.terasology.logic.characters.events.VerticalCollisionEvent;
+import org.terasology.logic.health.event.*;
+import org.terasology.logic.inventory.ItemComponent;
 import org.terasology.logic.players.event.OnPlayerRespawnedEvent;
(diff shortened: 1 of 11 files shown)


src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -21,20 +21,33 @@
...
-import org.terasology.logic.health.event.FullHealthEvent;
-import org.terasology.logic.health.event.OnHealedEvent;
+import org.terasology.logic.characters.CharacterMovementComponent;
+import org.terasology.logic.characters.CharacterSoundComponent;
+import org.terasology.logic.characters.CharacterSoundSystem;
+import org.terasology.logic.characters.MovementMode;
+import org.terasology.logic.characters.events.AttackEvent;
+import org.terasology.logic.characters.events.HorizontalCollisionEvent;
+import org.terasology.logic.characters.events.VerticalCollisionEvent;
+import org.terasology.logic.health.event.*;
```
syntaxi: Avoid star imports.
If IntelliJ is auto collapsing them then you can change the settings to increase the number of classes before it collapses to some high number like 999

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -140,16 +165,199 @@ private int calculateTotal(int base, TFloatList multipliers, TIntList modifiers)
...
+     *
+     * @param event Attack event sent on targetEntity.
+     * @param targetEntity The entity which is attacked.
+     */
+    @ReceiveEvent(components = HealthComponent.class, netFilter = RegisterMode.AUTHORITY)
+    public void onAttackEntity(AttackEvent event, EntityRef targetEntity) {
+        damageEntity(event, targetEntity);
+    }
+
+    static void damageEntity(AttackEvent event, EntityRef targetEntity) {
```
syntaxi: Why is this a separate static method to the above?

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -140,16 +165,199 @@ private int calculateTotal(int base, TFloatList multipliers, TIntList modifiers)
...
+
+    private void doDamage(EntityRef entity, int damageAmount, Prefab damageType, EntityRef instigator, EntityRef directCause) {
+        HealthComponent health = entity.getComponent(HealthComponent.class);
+        CharacterMovementComponent characterMovementComponent = entity.getComponent(CharacterMovementComponent.class);
+        boolean ghost = false;
+        if (characterMovementComponent != null) {
+            ghost = (characterMovementComponent.mode == MovementMode.GHOSTING);
+        }
+        if ((health != null) && !ghost) {
+            int damagedAmount = health.currentHealth - Math.max(health.currentHealth - damageAmount, 0);
```
syntaxi: This is overly complicated and the name is ambiguous.
Can't you replace this line with
```java
int damagedAmount = Math.min(health.currentHealth, damageAmount)
```
Further more based on that it makes sense to use `maxDamage` or `cappedDamage` as the name, eliminating the confusion between `damageAmount` and `damagedAmount`

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -140,16 +165,199 @@ private int calculateTotal(int base, TFloatList multipliers, TIntList modifiers)
...
+
+    /**
+     * Causes damage to entity when fallingDamageSpeedThreshold is breached.
+     *
+     * @param event VerticalCollisionEvent sent when falling speed threshold is crossed.
+     * @param entity The entity which is damaged due to falling.
+     */
+    @ReceiveEvent(components = {HealthComponent.class})
+    public void onLand(VerticalCollisionEvent event, EntityRef entity) {
+        HealthComponent health = entity.getComponent(HealthComponent.class);
```
syntaxi: If you are going to be using the component, just place it into the constructor to simplify things

ie,
```java
@ReceiveEvent
public void onLand(VerticalCollisionEvent event, EntityRef entity, HealthComponent health) {
    /* ... */
}
```

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -140,16 +165,199 @@ private int calculateTotal(int base, TFloatList multipliers, TIntList modifiers)
...
+     */
+    @ReceiveEvent(components = {HealthComponent.class})
+    public void onCrash(HorizontalCollisionEvent event, EntityRef entity) {
+        HealthComponent health = entity.getComponent(HealthComponent.class);
+
+        Vector3f vel = new Vector3f(event.getVelocity());
+        vel.y = 0;
+        float speed = vel.length();
+
+        if (speed > health.horizontalDamageSpeedThreshold) {
```
syntaxi: Parts of this method and the `onLand` method could likely be extracted into a single `highSpeedDamage` method

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -140,16 +165,199 @@ private int calculateTotal(int base, TFloatList multipliers, TIntList modifiers)
...
+     */
+    @ReceiveEvent
+    public void onCrash(HorizontalCollisionEvent event, EntityRef entity, CharacterSoundComponent characterSounds, HealthComponent healthComponent) {
+        Vector3f horizVelocity = new Vector3f(event.getVelocity());
+        horizVelocity.y = 0;
+        float velocity = horizVelocity.length();
+
+        if (velocity > healthComponent.horizontalDamageSpeedThreshold) {
+            if (characterSounds.lastSoundTime + CharacterSoundSystem.MIN_TIME < time.getGameTimeInMs()) {
+                StaticSound sound = random.nextItem(characterSounds.landingSounds);
```
syntaxi, line 358: Similarly, there are a few segments of code where sounds are played, which could be extracted into a common private method

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
syntaxi: This also applies to a similar section of the `doHeal` method

syntaxi (commented):
Tried to leave overall comments onto the code regardless of if you have written it.
Only minor things like simplifying and splitting code (reducing duplicates)

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
Cervator: Avoid wildcard import - configure IntelliJ (or whatever) to no collapse into wildcard, or at least not until 9999 imports or something crazy :-)

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -45,7 +58,18 @@
  * - BeforeHealEvent
  * - (HealthComponent saved)
  * - OnHealedEvent
- * - FullHealthEvent (if at full health)
+ * - OnFullyHealedEvent (if at full health)
```
Cervator: Maybe

```suggestion
 * - OnFullyHealedEvent (if healed to full)
```

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -140,16 +165,199 @@ private int calculateTotal(int base, TFloatList multipliers, TIntList modifiers)
 
     }
 
-    /** Handles DoHeal event to increase health of entity */
+    /**
+     * Handles DoHeal event to increase health of entity.
```
Cervator: Maybe tweak this with a code tag for `DoHealEvent` or even just avoid using the class name, instead referring to it generally "Event handler for increasing health ..."

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
Cervator: Missing access level? Also it might be worth throwing some bits of javadoc on some of these utility methods, even if not public :)

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -140,16 +165,199 @@ private int calculateTotal(int base, TFloatList multipliers, TIntList modifiers)
...
+        }
+    }
+
+    /**
+     * Causes damage to entity when fallingDamageSpeedThreshold is breached.
+     *
+     * @param event VerticalCollisionEvent sent when falling speed threshold is crossed.
+     * @param entity The entity which is damaged due to falling.
+     */
+    @ReceiveEvent(components = {HealthComponent.class})
```
Cervator: Since the `HealthComponent` is used inside this event handler you should be able to move it from the `@ReceiveEvent` filter into a method parameter. That way you don't need to separately retrieve it 👍

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -140,16 +165,199 @@ private int calculateTotal(int base, TFloatList multipliers, TIntList modifiers)
...
+        if (speed > health.fallingDamageSpeedThreshold) {
+            int damage = (int) ((speed - health.fallingDamageSpeedThreshold) * health.excessSpeedDamageMultiplier);
+            if (damage > 0) {
+                checkDamage(entity, damage, EngineDamageTypes.PHYSICAL.get(), EntityRef.NULL, EntityRef.NULL);
+            }
+        }
+    }
+
+    /**
+     * Inflicts damage to entity horizontalDamageSpeedThreshold is breached.
```
Cervator: Think there's an "if" missing here somewhere

src/main/java/org/terasology/logic/health/event/DamageSoundComponent.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.terasology.logic.health.event;
+
+import com.google.common.collect.Lists;
+import org.terasology.audio.StaticSound;
+import org.terasology.entitySystem.Component;
+
+import java.util.List;
+
+/**
```
Cervator, line 24: Blank javadoc - leftovers from when we did a bulk remove of `@author` tags ages ago. Feel free to remove them (or better yet fill them in!) wherever you see them

src/main/java/org/terasology/logic/health/event/HealthChangedEvent.java
```diff
@@ -37,7 +37,10 @@ public EntityRef getInstigator() {
     }
 
     /**
-     * @return The amount by which health changed. This is capped, so if the entity received 9999 damage and only had 10 health, it will be -10.
+     * Method to get the amount by which health has changed.
+     *
+     * @return The amount by which health changed. This is capped (by the implementing method),
+     *         so if the entity received 9999 damage and only had 10 health, it will be -10.
```
Cervator: It will _be_ `-10` or it will have `10` _subtracted_ from it?

src/test/java/org/terasology/logic/health/DamageEventTest.java
```diff
@@ -0,0 +1,148 @@
...
+package org.terasology.logic.health;
+
+import com.google.api.client.util.Sets;
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terasology.entitySystem.entity.EntityManager;
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.logic.health.event.*;
```
Cervator: More wildcard :-)

src/test/java/org/terasology/logic/health/DamageEventTest.java
```diff
@@ -0,0 +1,148 @@
...
+        player.addComponent(new PlayerCharacterComponent());
+        player.addComponent(healthComponent);
+
+        TestEventReceiver<BeforeDamagedEvent> receiver = new TestEventReceiver<>(getHostContext(),
+                BeforeDamagedEvent.class,
+                (event, entity) -> {
+                    event.add(5);
+                    event.multiply(2);
+                });
+        // Expected damage value is ( initial:10 + 5 ) * 2 = 30
```
Cervator, line 107: Might want to keep an eye out for `=` vs `==`

Cervator (changes requested):
Good stuff! Throwing in another round of fairly minor comments 👍

src/test/java/org/terasology/logic/health/DamageEventTest.java
e-aakash: I think this line is just calculation, so `=` should work. 
Will keep an eye out for them

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
e-aakash: I couldn't find any other methods using `damageEntity` method. So will move it into the `onAttackEntity` method.

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
e-aakash: Though this can be extracted, the `onCrash` only plays sound on the client (by sending PlaySoundEvent), but damage plays sound on both client and server.
So should the client sound way be extracted to a private method?

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
e-aakash: I did not check in the original Core/Health directory for usage. Turns out `BlockDamageAuthoritySystem` uses this method. [See here](https://github.com/e-aakash/Health/pull/1/files#diff-1a42c6b21c057e315502d39b4f22a5f6R225)
So I will have to either revert the commit or add it back in the next commit.

src/test/java/org/terasology/logic/health/DamageEventTest.java
Cervator: It is just math, true - I tend to always use the code style setup when it is in code even if it would be legit otherwise ... 😁

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -268,7 +276,7 @@ private void checkDamage(EntityRef entity, int amount, Prefab damageType, Entity
      * @param characterSounds Component having sound settings.
      */
     @ReceiveEvent
-    public void onDamaged(OnDamagedEvent event, EntityRef entity, CharacterSoundComponent characterSounds) {
+    public void onDamaged(OnDamagedEvent event, EntityRef entity, CharacterSoundComponent characterSounds)  {
```
Cervator, line 279: Super trivial but this ended up with two spaces instead of one for some reason

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -300,35 +308,31 @@ public void onDamaged(OnDamagedEvent event, EntityRef entity, CharacterSoundComp
...
         Vector3f vel = new Vector3f(event.getVelocity());
         vel.y = 0;
         float speed = vel.length();
 
-        if (speed > health.horizontalDamageSpeedThreshold) {
-            int damage = (int) ((speed - health.horizontalDamageSpeedThreshold) * health.excessSpeedDamageMultiplier);
+        highSpeedDamage(speed, entity, health.horizontalDamageSpeedThreshold, health.excessSpeedDamageMultiplier);
+    }
+
+    private void highSpeedDamage(float speed, EntityRef entity, float threshold, float damageMultiplier) {
```
Cervator, line 333: Nice utility method get! 👍

Cervator (approved):
Looks pretty solid 👍

src/test/java/org/terasology/logic/health/HealEventTest.java
```diff
@@ -20,11 +20,12 @@
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.terasology.engine.Time;
 import org.terasology.entitySystem.entity.EntityManager;
 import org.terasology.entitySystem.entity.EntityRef;
 import org.terasology.logic.health.event.BeforeHealEvent;
+import org.terasology.logic.health.event.DoDamageEvent;
```
Cervator, line 26: I think this is unused now

src/test/java/org/terasology/logic/health/HealEventTest.java
```diff
@@ -38,7 +39,6 @@
 public class HealEventTest extends ModuleTestingEnvironment {
 
     private EntityManager entityManager;
-    private Time time;
 
     private static final Logger logger = LoggerFactory.getLogger(HealEventTest.class);
```
Cervator, line 43: And this line could be moved above the `entityManager` to resolve a variable order issue

src/test/java/org/terasology/logic/health/HealEventTest.java
```diff
@@ -124,9 +123,27 @@ public void healModifyEventTest() {
...
                     event.multiply(2);
                 });
-        // expected = ( initial:10 + 10 - 5 ) * 2 = 30
+        // Expected value is ( initial:10 + 10 - 5 ) * 2 == 30
         player.send(new DoHealEvent(10));
         assertEquals(30, player.getComponent(HealthComponent.class).currentHealth);
     }
 
+    @Test
+    public void healNegativeTest() {
```
Cervator, line 132: Uh oh! This test actually fails for me! Expected 40, got 50.

Not worried about it though, merging anyway :-)

Cervator (approved):
Just a few more comments

PR #4, opened 2019-05-24
Block Damage systems

e-aakash:
Adds required assets and system for functioning of damage to blocks.

How to test:
1. Create a new game with `BuilderSampleGameplay` template with `Health` Module activated in advanced settings
2. Try to damage a block. See that the block takes damage before being destroyed.


iaronaraujo (approved):
Everything looks fine!

assets/prefabs/particleEffects/defaultBlockParticles.prefab
```diff
@@ -0,0 +1,31 @@
+{
+    "location": {},
+    "particleDataSprite": {
+        "texture": "core:fx_smoke"
```
syntaxi: This will need to be changed to "health:fx_smoke" and the file should be _duplicated_ to health because I'm fairly sure that there are other modules that use it.

src/main/java/org/terasology/logic/health/BlockDamageRenderer.java
```diff
@@ -0,0 +1,110 @@
...
+
+    private BlockSelectionRenderer blockSelectionRenderer;
+
+    @In
+    private EntityManager entityManager;
+
+    @Override
+    public void renderOverlay() {
+        if (blockSelectionRenderer == null) {
+            Texture texture = Assets.getTextureRegion("health:blockdamageeffects#1").get().getTexture();
```
e-aakash, line 51: Even though the [BlockDamageEffects](https://github.com/Terasology/Health/pull/4/files#diff-ec2d137ee6854b1fa1fbab10fb27efef) has been moved into the module, this line is creating `NoSuchElementException`.

deltas/engine/prefabs/test.prefab
```diff
@@ -0,0 +1,5 @@
+{
+    "RigidBody": {
+        "kinematic": false
```
syntaxi: Should this file still be here?

e-aakash: I don't think this is required, so will remove it.

iaronaraujo (approved):
The regen is working! It would be nice to open an issue to fix the regen speed later though

PR #5, opened 2019-05-30
Add Regenerate Logic

e-aakash:
Currently can't be tested directly, as only underlying events are changed.

How to test that regen still works:
1. Add `player.prefab` in Health/deltas/engine/prefabs/player from deltas in Core module.
2. Check out [PR#3677](https://github.com/MovingBlocks/Terasology/pull/3677) branch in Terasology repo.
3. Activate Health module in a new game, execute console command `damage 30`. 
4. See that regen starts after 10 secs and full health is attained at 20 secs after the damage is inflicted.

### DIFF
--- a/src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
@@ -42,12 +42,14 @@ import org.terasology.logic.characters.events.HorizontalCollisionEvent;
 import org.terasology.logic.characters.events.VerticalCollisionEvent;
 import org.terasology.logic.health.event.BeforeDamagedEvent;
 import org.terasology.logic.health.event.BeforeHealEvent;
+import org.terasology.logic.health.event.BeforeRegenEvent;
 import org.terasology.logic.health.event.DamageSoundComponent;
 import org.terasology.logic.health.event.DoDamageEvent;
 import org.terasology.logic.health.event.DoHealEvent;
 import org.terasology.logic.health.event.OnDamagedEvent;
 import org.terasology.logic.health.event.OnFullyHealedEvent;
 import org.terasology.logic.health.event.OnHealedEvent;
+import org.terasology.logic.health.event.OnRegenedEvent;
 import org.terasology.logic.inventory.ItemComponent;
 import org.terasology.logic.players.event.OnPlayerRespawnedEvent;
 import org.terasology.math.TeraMath;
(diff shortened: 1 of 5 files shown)


src/main/java/org/terasology/logic/health/event/RegenerateEvent.java
```diff
@@ -0,0 +1,63 @@
+/*
```
e-aakash, line 1: Currently, this event is not used. After the discussion on how to do regen is concluded, will update/use this in the authority system.

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -119,13 +127,36 @@ private int regenerateHealth(HealthComponent health, int healAmount) {
...
+                    doDamage(entity, -modifiedAmount, EngineDamageTypes.HEALING.get(), entity, EntityRef.NULL);
+                }
+            }
             entity.saveComponent(health);
         }
     }
 
+    private void doRegenerate(EntityRef entity, int healAmount, EntityRef instigator) {
+        HealthComponent health = entity.getComponent(HealthComponent.class);
+        if (health != null) {
```
e-aakash, line 147: These (4 lines) can be taken out as a helper function for both Regen and Restore. Will refactor this after also adding restore system.

e-aakash (commented):
Added some clarification comments.

Cervator (approved):
Been looking at the original code, some related bits & history, and chat on Slack. Rather than line-comment for this one I think I'll instead outline some higher level desires I'd personally go for and see what you think. Consider this critique aimed more at the origin code than your changes :-)

* This class is _ancient_ - 2012, with some bits showing their age
* I think it does too much, and with regen it is doing even more. I think cloning the regen structure in this PR is fine, but I would split out regen + health + damage into each their own system, possibly with a little bit of static utility code floating around somewhere
* Why does it do so much? Probably because it is a single `UpdateSubscriberSystem` meaning that one ticking `update` method could go through _everything_ with a `HealthComponent` then process in multiple ways (heal/harm)
* If so then by splitting apart things we don't want to end up with 3 separate `update` ticks - but we _could_ use the scheduler instead, have _no_ `update` ticks not already done (the scheduler itself), and avoid the `entityManager.getEntitiesWith` entirely.

Not that health processing is likely much of a performance hog at present but that sort of change should massively reduce the amount of process + split out things into smaller easier to digest pieces. How does that sound?

I don't mind merging this as is so we can iterate in a new direction in a new PR :-)

For `RegenerationAuthoritySystem` in AlterationEffects it already kinda looks like a piece pulled out of `HealthAuthoritySystem` complete with its own `update` (throttled at least) - maybe the two approaches could be unified in Health and AE just made to depend on it instead?

Cervator:
Oh, and this tests out, so am merging 👍

PR #8, opened 2019-06-06
Add regen component

e-aakash:
Currently made changes so that existing system (Health, Damage) don't break.

Will add scheduled regen using this component.

### DIFF
--- a/src/main/java/org/terasology/logic/health/RegenComponent.java
+++ b/src/main/java/org/terasology/logic/health/RegenComponent.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2019 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.logic.health;
+
+import org.terasology.entitySystem.Component;
+import org.terasology.network.Replicate;
+
+public class RegenComponent implements Component {
+    /** Amount of health restored in each regeneration tick. */
+    @Replicate
+    public float regenRate;
+
+    /** Time delay before regeneration starts. */
+    @Replicate
+    public float waitBeforeRegen;
+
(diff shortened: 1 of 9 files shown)


src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -0,0 +1,83 @@
...
+package org.terasology.logic.health;
+
+import org.terasology.entitySystem.Component;
+import org.terasology.math.TeraMath;
+import org.terasology.network.Replicate;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class RegenComponent implements Component {
```
Cervator, line 25: I wonder if we should fully specify this as `RegeneratingComponent` - would be clearer (entity is actively regenerating) and leave more potential space if a later system want to mark entities with something like `RegenerableComponent`

src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -0,0 +1,83 @@
...
+    private Map<String, Float> regenValue = new HashMap<>();
+    private Map<String, Long> regenEndTime = new HashMap<>();
+    public long lowestEndTime;
```
Cervator, line 28: I wonder if the wording could be improved from "lowest" here - to me at least it doesn't seem super intuitive vs the concepts of starting and ending things. `soonest` instead perhaps?

src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -0,0 +1,83 @@
...
+    public boolean isEmpty() {
+        return regenValue.isEmpty();
+    }
+
+    public int getRegenValue() {
+        float totalValue = 0;
+        for (float value: regenValue.values()) {
+            totalValue += value;
+        }
+        if (totalValue < 0) {
```
Cervator, line 77: Maybe I'm mistaken but I have a feeling there should be _some_ sort of utility method either in Java Math or our Math lib that should be able to take a number and set it to 0 if it is indeed below 0 (or any arbitrary number). Pretty trivial though, but maybe if there's something obvious?

Cervator (approved):
Tested out locally and reviewed - looks pretty good overall I'd say, although it might be good with another set of eyes just in case. Found #9 but didn't expect multiplayer to be flawless yet. Going to merge as-is since I resolved some conflicts 👍

src/main/java/org/terasology/logic/health/RegenComponent.java
syntaxi: `Math.max(totalValue, 0);`

syntaxi: `Math.max(totalValue, 0);`?

Issue #9, opened 2019-06-10
Concurrent modification server crash after PR #8


Cervator:
Probably related to two paths hitting the same map in the regeneration scheduler

Scenario: Server/client hosting a BuilderSampleGameplay world, 2nd client joins. Beat randomly on a few blocks and observe yep they don't regen right now, that's another known issue.

Host beat on 2nd client via fisticuffs for a few seconds, observed regen happen fine for 2nd client.

2nd client beat on host for several more seconds, observed regen starting fine but then crashed the server:

```
21:15:18.188 [main] ERROR o.t.e.e.internal.PojoEntityManager - Adding a component (class org.terasology.logic.health.RegenComponent) over an existing component for entity 31
21:15:18.188 [main] WARN  o.t.logic.health.RegenComponent - Regen added baseRegen
21:15:18.188 [main] WARN  o.t.logic.health.RegenComponent - Regen added baseRegen
21:15:18.241 [main] WARN  o.t.logic.health.RegenComponent - regenerating 3
21:15:18.241 [main] ERROR o.terasology.engine.TerasologyEngine - Uncaught exception, attempting clean game shutdown
java.util.ConcurrentModificationException: null
	at java.util.TreeMap$PrivateEntryIterator.nextEntry(TreeMap.java:1211)
	at java.util.TreeMap$EntryIterator.next(TreeMap.java:1247)
	at java.util.TreeMap$EntryIterator.next(TreeMap.java:1242)
	at com.google.common.collect.AbstractMapBasedMultimap$Itr.next(AbstractMapBasedMultimap.java:1177)
	at org.terasology.logic.health.RegenAuthoritySystem.regenerate(RegenAuthoritySystem.java:102)
	at org.terasology.logic.health.RegenAuthoritySystem.invokeRegenOperations(RegenAuthoritySystem.java:98)
	at org.terasology.logic.health.RegenAuthoritySystem.update(RegenAuthoritySystem.java:70)
	at org.terasology.engine.modes.StateIngame.update(StateIngame.java:171)
	at org.terasology.engine.TerasologyEngine.tick(TerasologyEngine.java:458)
	at org.terasology.engine.TerasologyEngine.mainLoop(TerasologyEngine.java:421)
	at org.terasology.engine.TerasologyEngine.run(TerasologyEngine.java:397)
	at org.terasology.engine.Terasology.main(Terasology.java:156)
[...]

PR #10, opened 2019-06-11
Handle less than 1 regen and removing Regen logic

e-aakash:
Solves https://github.com/Terasology/Health/issues/9

How to test:
1. Create a new game with BSG and Health module
2. Attack more than one block
3. They get properly regenerated, without throwing an exception

How to test multiplayer
1. Create a server
2. Join using another client
3. Attack each other
4. Both get regenerated

### DIFF
--- a/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
@@ -29,7 +29,9 @@ import org.terasology.entitySystem.systems.RegisterMode;
 import org.terasology.entitySystem.systems.RegisterSystem;
 import org.terasology.entitySystem.systems.UpdateSubscriberSystem;
 import org.terasology.logic.health.event.ActivateRegenEvent;
+import org.terasology.logic.health.event.DeactivateRegenEvent;
 import org.terasology.logic.health.event.OnFullyHealedEvent;
+import org.terasology.protobuf.EntityData;
 import org.terasology.registry.In;
 
 import java.util.Iterator;
@@ -38,22 +40,24 @@ import java.util.List;
 
 /**
  * This system handles the natural regeneration of entities with HealthComponent.
- *
  */
 @RegisterSystem(RegisterMode.AUTHORITY)
 public class RegenAuthoritySystem extends BaseComponentSystem implements UpdateSubscriberSystem {
 
     private static final Logger logger = LoggerFactory.getLogger(RegenComponent.class);
+    /**
+     * Integer storing when to check each effect.
+     */
+    private static final int CHECK_INTERVAL = 200;
+
+    /**
+     * Integer storing when last time entities were regenerated.
+     */
+    private static long lastTick = 0;
 
     // Stores when next to check for new value of regen, contains only entities which are being regenerated
     private SortedSetMultimap<Long, EntityRef> regenSortedByTime = TreeMultimap.create(Ordering.natural(),
             Ordering.arbitrary());
-
-    /** Integer storing when to check each effect. */
-    private static final int CHECK_INTERVAL = 200;
-
-    /** Integer storing when to check each effect. */
-    private static long lastTick = 0;
 
     @In
     private Time time;
@@ -109,7 +113,7 @@ public class RegenAuthoritySystem extends BaseComponentSystem implements UpdateS
 
     private void regenerate(long currentTime) {
         List<Long> regenToBeRemoved = new LinkedList<>();
-        for (EntityRef entity: regenSortedByTime.values()) {
+        for (EntityRef entity : regenSortedByTime.values()) {
             RegenComponent regen = entity.getComponent(RegenComponent.class);
             HealthComponent health = entity.getComponent(HealthComponent.class);
             if (health.nextRegenTick < currentTime) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -67,10 +74,14 @@
     public void update(float delta) {
         final long currentTime = time.getGameTimeInMs();
         // Execute regen schedule
-        invokeRegenOperations(currentTime);
+        if (currentTime > lastTick + CHECK_INTERVAL) {
+            invokeRegenOperations(currentTime);
+            lastTick = currentTime;
```
syntaxi: _very_ minor point but to reduce the work done every frame you should do the addition of `CHECK_INTERVAL` here and not in the if condition

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -95,39 +107,48 @@ private void invokeRegenOperations(long currentWorldTime) {
...
                 if (health.currentHealth >= health.maxHealth) {
-                    regenSortedByTime.remove(regen.getLowestEndTime(), entity);
+                    regenToBeRemoved.add(regen.getLowestEndTime());
                     entity.removeComponent(RegenComponent.class);
                     entity.send(new OnFullyHealedEvent(entity));
                 }
                 entity.saveComponent(health);
             }
         }
+        for (Long endTime : regenToBeRemoved) {
```
syntaxi, line 135: If you feel like being fancy you _can_ use a lambda here.
Not suggesting that you do do this, leaving it as a loop is clear. But just giving you more info.

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -144,4 +165,20 @@ private void addRegenToScheduler(ActivateRegenEvent event, EntityRef entity, Reg
         regenSortedByTime.put(regen.getLowestEndTime(), entity);
     }
 
+    @ReceiveEvent
+    public void onRegenRemoved(DeactivateRegenEvent event, EntityRef entity, HealthComponent health,
+                               RegenComponent regen) {
+        regenSortedByTime.remove(regen.getLowestEndTime(), entity);
+        if (event.id.equals("all")) {
```
syntaxi: Can we get these strings (`"all"` and `"baseRegen"`) as _public_ static fields.

src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -70,14 +71,17 @@ public boolean isEmpty() {
...
 
     public int getRegenValue() {
-        float totalValue = 0;
+        float totalValue = remainder;
         for (float value: regenValue.values()) {
             totalValue += value;
         }
-        if (totalValue < 0) {
-            totalValue = 0;
+        remainder = Math.min(1, remainder);
```
syntaxi: You can _actually_ do
```java
remainder = totalValue % 1 //modulo 1
```
To get the remainder when dividing by 1.
This saves you having to floor to int and then subtract to get the return value

src/main/java/org/terasology/logic/health/event/DeactivateRegenEvent.java
```diff
@@ -0,0 +1,30 @@
...
+package org.terasology.logic.health.event;
+
+import org.terasology.entitySystem.event.Event;
+
+public class DeactivateRegenEvent implements Event {
+    public String id;
+
+    public DeactivateRegenEvent() {
+        id = "baseRegen";
+    }
```
syntaxi, line 27: Should point to the public static field I mentioned above

syntaxi (commented):
Structure of the system looks good, only small issues here and there.

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -95,53 +111,78 @@ private void invokeRegenOperations(long currentWorldTime) {
...
     }
 
     private void regenerate(long currentTime) {
-        for (EntityRef entity: regenSortedByTime.values()) {
+        List<Long> regenToBeRemoved = new LinkedList<>();
+        for (EntityRef entity : regenSortedByTime.values()) {
             RegenComponent regen = entity.getComponent(RegenComponent.class);
             HealthComponent health = entity.getComponent(HealthComponent.class);
-            if (health.nextRegenTick < currentTime) {
+            if (health != null && health.nextRegenTick < currentTime) {
```
e-aakash, line 123: Added null check because NPE is thrown when blocks are destroyed.

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -33,6 +33,7 @@
 import org.terasology.logic.health.event.OnFullyHealedEvent;
 import org.terasology.protobuf.EntityData;
 import org.terasology.registry.In;
+import org.terasology.world.generation.BaseFacetedWorldGenerator;
```
syntaxi, line 36: Accidental import?

syntaxi (approved):

Cervator:
Tested out and merged :-)

I added a tiny commit to remove a few unused bits.

PR #12, opened 2019-06-16
Code quality tweaks 

e-aakash:
Remove logs and unused events.
Update tests for regeneration


src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -22,53 +22,33 @@
...
         long result = Long.MAX_VALUE;
         for (long value : regenEndTime.values()) {
             result = Math.min(result, value);
         }
-        return lowestEndTime;
-    }
-
-    public long getLowestEndTime() {
-        return lowestEndTime;
+        return soonestEndTime;
```
syntaxi: You need to do something with result here

src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -22,53 +22,33 @@
 import java.util.Map;
 
 public class RegenComponent implements Component {
-    public long lowestEndTime;
-    private Map<String, Float> regenValue = new HashMap<>();
-    private Map<String, Long> regenEndTime = new HashMap<>();
-    private float remainder;
+    public long soonestEndTime;
+    public Map<String, Float> regenValue = new HashMap<>();
+    public Map<String, Long> regenEndTime = new HashMap<>();
```
syntaxi, line 27: If you can find a way, make this a map sorted on the `long` value. It will _greatly_ improve performance because at present to find min, & to remove below a specific amount then you can quickly and easily do this. It does make removing by id harder, but this is not as often needed.

syntaxi (commented):
Few points

Cervator:
Closing as superseded by #16, which is now merged 👍

PR #13, opened 2019-06-16
Removal of RegenComponent with checks

### LINKED ISSUE #11: RegenComponent is removed at full health
If a player drinks a regen potion, it activates regen by `ActivateRegenEvent()`. But the regen component, and hence the regeneration information is lost if the player is at full health.

e-aakash:
Closes #11 

`RegenComponent` is removed if only base regeneration is active.

### DIFF
--- a/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
@@ -124,7 +124,9 @@ public class RegenAuthoritySystem extends BaseComponentSystem implements UpdateS
                 health.nextRegenTick = currentTime + 1000;
                 if (health.currentHealth >= health.maxHealth) {
                     regenToBeRemoved.add(regen.getLowestEndTime());
-                    entity.removeComponent(RegenComponent.class);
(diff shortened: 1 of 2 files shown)


src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -124,7 +124,9 @@ private void regenerate(long currentTime) {
                 health.nextRegenTick = currentTime + 1000;
                 if (health.currentHealth >= health.maxHealth) {
                     regenToBeRemoved.add(regen.getLowestEndTime());
-                    entity.removeComponent(RegenComponent.class);
+                    if (regen.hasBaseRegenOnly()) {
+                        entity.removeComponent(RegenComponent.class);
+                    }
                     entity.send(new OnFullyHealedEvent(entity));
```
e-aakash, line 130: This will not lead to multiple `OnFullyHealedEvent` as the entity is removed from the scheduler on reaching full health ([Line 126](https://github.com/Terasology/Health/pull/13/files#diff-7de79d955d70c98aa3d1ff0a21c48faeR126))

PR #15, opened 2019-06-19
Restoration Logic

e-aakash:
Adds new events and logic for handling restoration.

### DIFF
--- a/src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
@@ -15,10 +15,49 @@
  */
 package org.terasology.logic.health;
 
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.ReceiveEvent;
 import org.terasology.entitySystem.systems.BaseComponentSystem;
 import org.terasology.entitySystem.systems.RegisterMode;
 import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.logic.health.event.BeforeRestoreEvent;
+import org.terasology.logic.health.event.DoDamageEvent;
+import org.terasology.logic.health.event.DoRestoreEvent;
+import org.terasology.logic.health.event.OnFullyHealedEvent;
+import org.terasology.logic.health.event.OnRestoredEvent;
+import org.terasology.logic.players.event.OnPlayerRespawnedEvent;
+import org.terasology.math.TeraMath;
 
 @RegisterSystem(RegisterMode.AUTHORITY)
 public class RestorationAuthoritySystem extends BaseComponentSystem {
+
+    @ReceiveEvent
+    public void onRestore(DoRestoreEvent event, EntityRef entity, HealthComponent health) {
+        BeforeRestoreEvent beforeRestoreEvent = entity.send(new BeforeRestoreEvent(event.getAmount(), event.getEntity()));
+        if (!beforeRestoreEvent.isConsumed()) {
+            int modifiedRestoreAmount = TeraMath.floorToInt(beforeRestoreEvent.getResultValue());
+            if (modifiedRestoreAmount > 0) {
+                restore(entity, health, modifiedRestoreAmount);
+            } else {
+                entity.send(new DoDamageEvent(-modifiedRestoreAmount));
+            }
+        }
+    }
+
+    private void restore(EntityRef entity, HealthComponent health, int restoreAmount) {
+        int cappedHealth = Math.min(health.maxHealth, health.currentHealth + restoreAmount);
+        int cappedRestoreAmount = cappedHealth - health.currentHealth;
+        health.currentHealth = cappedHealth;
+        entity.saveComponent(health);
+        entity.send(new OnRestoredEvent(cappedRestoreAmount, entity));
+        if (cappedHealth == health.maxHealth) {
+            entity.send(new OnFullyHealedEvent(entity));
+        }
+    }
+
+    @ReceiveEvent
+    public void onRespawn(OnPlayerRespawnedEvent event, EntityRef entity, HealthComponent healthComponent) {
+        healthComponent.currentHealth = healthComponent.maxHealth;
+        entity.saveComponent(healthComponent);
+    }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
```diff
@@ -0,0 +1,63 @@
...
+import org.terasology.math.TeraMath;
+
+@RegisterSystem(RegisterMode.AUTHORITY)
+public class RestorationAuthoritySystem extends BaseComponentSystem {
+
+    @ReceiveEvent
+    public void onRestore(DoRestoreEvent event, EntityRef entity, HealthComponent health) {
+        BeforeRestoreEvent beforeRestoreEvent = entity.send(new BeforeRestoreEvent(event.getAmount(), event.getEntity()));
+        if (!beforeRestoreEvent.isConsumed()) {
+            int modifiedRestoreAmount = TeraMath.floorToInt(beforeRestoreEvent.getResultValue());
```
syntaxi, line 38: Where's this `getResultValue` method located? because I can't see it in the `BeforeRestoreEvent`.
Additionally, rather than flooring to an int, just have the value as an int

src/main/java/org/terasology/logic/health/event/DoRestoreEvent.java
```diff
@@ -0,0 +1,63 @@
...
+    public DoRestoreEvent(int amount) {
+        this(amount, EntityRef.NULL);
+    }
+
+    /**
+     * Constructor for event, with amount and entity specified.
+     * @param amount            The amount of restoration.
+     * @param restoreEntity     The entity which is being restored.
+     */
+    public DoRestoreEvent(int amount, EntityRef restoreEntity) {
```
syntaxi: Shouldn't this event just be sent against the entity to restore?

syntaxi (commented):
Two points at the moment.
Not much in the way of logic to review but the structure looks good to me

src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
e-aakash: It is present in the `AbstractConsumableValueModifiableEvent `, which is extended by `BeforeRestoreEvent`. Since it has a get result method (which returns a float), I used it, adding a floor to int to make it useful for our case. 
Can add a result method in `BeforeRestoreEvent` itself, if needed,

src/main/java/org/terasology/logic/health/event/DoRestoreEvent.java
e-aakash: Will remove the `EntityRef`, as it is not needed.

src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
```diff
@@ -0,0 +1,63 @@
...
+import org.terasology.logic.health.event.BeforeRestoreEvent;
+import org.terasology.logic.health.event.DoDamageEvent;
+import org.terasology.logic.health.event.DoRestoreEvent;
+import org.terasology.logic.health.event.OnFullyHealedEvent;
+import org.terasology.logic.health.event.OnRestoredEvent;
+import org.terasology.logic.players.event.OnPlayerRespawnedEvent;
+import org.terasology.math.TeraMath;
+
+@RegisterSystem(RegisterMode.AUTHORITY)
+public class RestorationAuthoritySystem extends BaseComponentSystem {
```
Cervator, line 32: Class-level Javadoc would be nice here, maybe to highlight how restoration is intended to differ from regeneration (sort of a sibling system)

src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @ReceiveEvent
+    public void onRestore(DoRestoreEvent event, EntityRef entity, HealthComponent health) {
+        BeforeRestoreEvent beforeRestoreEvent = entity.send(new BeforeRestoreEvent(event.getAmount(), entity));
+        if (!beforeRestoreEvent.isConsumed()) {
+            int modifiedRestoreAmount = TeraMath.floorToInt(beforeRestoreEvent.getResultValue());
+            if (modifiedRestoreAmount > 0) {
+                restore(entity, health, modifiedRestoreAmount);
+            } else {
+                entity.send(new DoDamageEvent(-modifiedRestoreAmount));
```
Cervator, line 42: Comments to briefly highlight how this and the prior restore line are intended to work would probably also be helpful for a viewer who is just glancing through the code and goes "wait what" :-)

src/main/java/org/terasology/logic/health/event/BeforeRestoreEvent.java
```diff
@@ -0,0 +1,49 @@
...
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.entitySystem.event.AbstractConsumableValueModifiableEvent;
+
+/**
+ * This event is sent to an entity to allow modification and cancellation of restoration.
+ *
+ * Modifications are accumulated as modifiers (additions), multipliers and postModifiers (additions after multipliers).
+ */
+public class BeforeRestoreEvent extends AbstractConsumableValueModifiableEvent {
+    /** The entity which is being regenerated. */
```
Cervator, line 27: "restored" probably ;-)

src/main/java/org/terasology/logic/health/event/OnRestoredEvent.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terasology.logic.health.event;
+
+import org.terasology.entitySystem.entity.EntityRef;
+
+public class OnRestoredEvent extends HealthChangedEvent {
```
Cervator, line 20: In line with the other two events having javadoc it would be nice if this one did too 👍

Cervator (approved):
Am fine with the PR, pinging a few minor line issues and improvements that can be done next PR 👍

PR #16, opened 2019-06-20
For solving the multi block non healing bug

e-aakash:
https://github.com/Terasology/Health/pull/12 rebased on master.
Added some loggers for finding whats going wrong

How to test:
1. Start a game with BSG and Health module.
2. Hit many blocks consecutively.
3. See that all of them regenerate, and not any gets "stuck" in damaged state.


Cervator:
Tests out! Ran all the unit tests, tried BSG in both single player and multiplayer, everything seems to be working. Nicely done :-)

One observation, which I actually favor, at least personally: regeneration seems a little erratic in timing, which I think adds a bit of spice instead of mechanical on-the-dot accuracy. Can see if you damage a large area of blocks in a row or if you heavily damage a player then wait and time how fast each half-heart returns.

PR #17, opened 2019-06-22
Add javadoc to restore events and RestorationAuthoritySystem

e-aakash:
No major code changes

### DIFF
--- a/src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
@@ -28,6 +28,17 @@ import org.terasology.logic.health.event.OnRestoredEvent;
 import org.terasology.logic.players.event.OnPlayerRespawnedEvent;
 import org.terasology.math.TeraMath;
 
+/**
+ * This system takes care of restoration of entities with HealthComponent.
+ * To increase the health of an entity, send DoRestoreEvent
+ * <p>
+ * Logic flow for restoration:
+ * - DoRestoreEvent
+ * - BeforeRestoreEvent
+ * - (HealthComponent saved)
+ * - OnRestoredEvent
+ * - OnFullyHealedEvent (if healed to full health)
+ */
 @RegisterSystem(RegisterMode.AUTHORITY)
 public class RestorationAuthoritySystem extends BaseComponentSystem {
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/terasology/logic/health/event/DoRestoreEvent.java
```diff
@@ -23,13 +24,20 @@
 public class DoRestoreEvent implements Event {
     /** The amount of health points being restored. */
     private int amount;
+    /** The instigator of restoration */
```
Cervator: Tiny suggestion: rather than use `instigator` which is already the variable name use a synonym to increase the odds that a reader will understand (instigator could be a somewhat rare word)

```suggestion
    /** The entity that caused the restoration */
```

src/main/java/org/terasology/logic/health/RestorationAuthoritySystem.java
```diff
@@ -28,6 +28,17 @@
 import org.terasology.logic.players.event.OnPlayerRespawnedEvent;
 import org.terasology.math.TeraMath;
 
+/**
+ * This system takes care of restoration of entities with HealthComponent.
+ * To increase the health of an entity, send DoRestoreEvent
+ * <p>
+ * Logic flow for restoration:
+ * - DoRestoreEvent
```
Cervator, line 36: This is too big to consider as part of this PR (since the same word is in use many other places) but I wonder if `do` is a good term in general for the starting phase of an event chain like this one. "Do" to me very much means that the thing is happening right now. Maybe "Perform" would give a better impression that it is more like an action that is being performed, almost like a stage play might be going through several phases to arrive at an action :-)

This could be overly subjective as interpreting these kinds of words may differ subtly between individuals.

src/main/java/org/terasology/logic/health/event/DoRestoreEvent.java
```diff
@@ -39,4 +47,12 @@ public DoRestoreEvent(int amount) {
     public int getAmount() {
         return amount;
     }
+
+    /**
+     * Method to get the instigator of restoration.
```
Cervator: Generally I'd excuse setters & getters from Javadoc unless there is any potential for confusion somehow.

src/main/java/org/terasology/logic/health/event/OnRestoredEvent.java
```diff
@@ -17,14 +17,29 @@
...
+ * This event is sent after the entity is restored. Final event of restoration logic.
+ */
 public class OnRestoredEvent extends HealthChangedEvent {
+    /**
+     * The amount by which the entity is restored.
+     */
     private int amount;
 
+    /**
+     * Constructor for creating the event.
```
Cervator: Another edge case: should plain constructors have javadoc? In this case maybe if the wording is a slight bit more meaningful (we already know by the constructors presence that it will create the event, so the text doesn't really add anything). I'm not sure if there's really any other meaning to add in this case.

Cervator (approved):
Nice and easy to review and approve! Left a few comments anyway but they're more about trivial wording issues than anything blocking a merge.

PR #22, opened 2019-06-26
Update system to handle reload

### LINKED ISSUE #14: Save and reload stops ongoing regen
When a block is damaged, and the game is reloaded, the block remains at same "damage state". 

The cause might be that scheduler information is lost on leaving the game. 

Possible solution: Add a `postBegin()` to schedule regen

e-aakash:
Fixes https://github.com/Terasology/Health/issues/14

How to test:
1. Make sure regen happens in-game (console command `damage 10`, player regenerates after some time, and blocks regenerate after hitting)
2. Damage some blocks, open pause menu (to preserve block in the damaged state and wait for save), reload the game. Blocks get regenerated.

### DIFF
--- a/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
@@ -149,7 +149,7 @@ public class RegenAuthoritySystem extends BaseComponentSystem implements UpdateS
         List<String> toBeRemoved = new LinkedList<>();
         for (String id : regen.regenEndTime.keySet()) {
             endTime = regen.regenEndTime.get(id);
-            if ((endTime != -1) && (endTime <= currentTime)) {
+            if (endTime != -1 && endTime <= currentTime) {
                 toBeRemoved.add(id);
             }
         }
@@ -194,7 +194,6 @@ public class RegenAuthoritySystem extends BaseComponentSystem implements UpdateS
 
     @ReceiveEvent
     public void onRegenComponentAdded(OnActivatedComponent event, EntityRef entity, RegenComponent regen) {
-        logger.warn("found regen getting added on " + entity.getId() + " " + regen.soonestEndTime + " " + time.getGameTimeInMs());
         regenSortedByTime.put(regen.soonestEndTime, entity);
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -140,15 +144,19 @@ private void regenerate(long currentTime) {
...
-        long endTime;
+    private void removeCompleted(Long currentTime, RegenComponent regen) {
+        Long endTime;
+        List<String> toBeRemoved = new LinkedList<>();
         for (String id : regen.regenEndTime.keySet()) {
             endTime = regen.regenEndTime.get(id);
-            if (endTime <= currentTime) {
-                regen.regenEndTime.remove(id);
-                regen.regenValue.remove(id);
+            if ((endTime != -1) && (endTime <= currentTime)) {
```
Cervator: I don't think you need the extra `( )` here around each condition

Cervator (approved):
This seems fine, just one trivial code quality tweak you could do - but the weird part is that I can't replicate the original issue anymore. At very latest `master` for Health reloads already don't break the regen, even in multiplayer. Huh?

Can verify that it also seems to work fine after merging this. Did it get fixed coincidentally already? I figure this method might be better/cleaner anyway so hey ... :-)

Cervator:
Never mind, I wasn't testing it right. Current state doesn't work, this does 😁

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -159,28 +167,34 @@ public void onRegenAdded(ActivateRegenEvent event, EntityRef entity, RegenCompon
...
             regen.addRegen(WAIT, -health.regenRate,
                     time.getGameTimeInMs() + (long) (health.waitBeforeRegen * 1000));
         } else {
             regen.addRegen(event.id, event.value, time.getGameTimeInMs() + (long) (event.endTime * 1000));
         }
+    }
+
+    @ReceiveEvent
+    public void onRegenComponentAdded(OnActivatedComponent event, EntityRef entity, RegenComponent regen) {
+        logger.warn("found regen getting added on " + entity.getId() + " " + regen.soonestEndTime + " " + time.getGameTimeInMs());
```
syntaxi: Needs to be removed before merge ;)

PR #25, opened 2019-06-28
Add event to restore full health

### LINKED ISSUE #19: Instantly heal player to full health
Some event should be provided for other systems to heal an entity to full health. 
Currently, other systems increase the health value by huge amount to do the same. Providing with an event for it will improve intersystem interaction, as they will not have to directly access `HealthComponent`.

e-aakash:
No method to test in-game. 
Run the test to confirm its working.

Closes https://github.com/Terasology/Health/issues/19


syntaxi:
Tests run fine, and the change is small & seems fine to me.

Issue #21, opened 2019-06-26
Base Regen handling


e-aakash:
It would be better if `baseRegen` with 0 value is not added to the scheduler.

e-aakash:
Closed by https://github.com/Terasology/Health/pull/27

PR #27, opened 2019-07-03
Handle 0 regen value

e-aakash:
Does not add regeneration to the component if `regenValue` is zero.
`RegenTest.regenTest()` work in isolation, but fails when all tests in `RegenTest.java` are run

### DIFF
--- a/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
@@ -159,10 +159,12 @@ public class RegenAuthoritySystem extends BaseComponentSystem implements UpdateS
     @ReceiveEvent
     public void onRegenAdded(ActivateRegenEvent event, EntityRef entity, RegenComponent regen,
                              HealthComponent health) {
-        // Remove previous scheduled regen, new will be added by addRegenToScheduler()
-        regenSortedByTime.remove(regen.soonestEndTime, entity);
-        addRegenToScheduler(event, entity, regen, health);
-        regenSortedByTime.put(regen.soonestEndTime, entity);
+        if (event.value != 0) {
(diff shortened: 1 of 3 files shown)


Cervator:
Protip: For testing a change like this it would be very handy to inline a couple example prefabs / edits to existing ones + expected state before and after applying the PR :-)

Cervator:
I merged this locally and played around with it for a bit. I think I see _what_ is wrong when running all tests together,, but not _why_

At the beginning of each test game time is just a few ms. When run individually all tests work fine. The `multipleRegenTest` runs first when grouped, and ticks for 6.2 seconds. When `regenTest` runs next game time _says_ it is back to just a few ms, but regen doesn't trigger. If you let it tick for its desired value plus 6.2s (or comment out the `runWhile` in the other test) then it works fine.

Even though time says it resets, _something_ is still 6.2s along when the `regenTest` begins.

I coincidentally just merged https://github.com/MovingBlocks/TerasologyLauncher/pull/449 which dealt with another _leaky_ test that affected the environment for a later test. The fix there was adding a test handler to reset the environment after the test was done. I'm not sure if we can target something else to truly reset time, or might be looking at a bug in the MTE? 

Also, could we simply just add a 4th test that exercises an attempt to add a regen with a value of 0? That'd be a nice way to validate this PR 👍 

There's also a small conflict now. I fixed it locally (hopefully) by just wrapping `if (value != 0)` around changed code, but am not sure if the `removeRegen(id);` should be inside or outside the if. I'm thinking inside - otherwise having a pointless regen effect applied with a strength of 0 would stop existing (valid) regeneration.

Finally: Does this fix #21 when merged?

e-aakash:
Yes, the `removeRegen(id)` would be inside the if, and this pr does fix issue 21

Added a test for this pr

PR #28, opened 2019-07-03
Convert regenEndTime to sorted map

### LINKED ISSUE #26: Use sorted map for regenEndTime
Converting [regenEndTime](https://github.com/Terasology/Health/blob/master/src/main/java/org/terasology/logic/health/RegenComponent.java#L29) map to sorted map removes the need for `soonestEndTime` and can improve performance during removal of completed regen.

e-aakash:
Fixes https://github.com/Terasology/Health/issues/26

How to test:
Make sure regen happens in-game

### DIFF
--- a/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
@@ -140,17 +140,24 @@ public class RegenAuthoritySystem extends BaseComponentSystem implements UpdateS
     }
 
     private void removeCompleted(Long currentTime, RegenComponent regen) {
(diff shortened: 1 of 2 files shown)


Cervator:
Heh, whoops, was curious what would happen if I merged a PR branch connected with a Draft - turns out the commit gets merged but the PR gets stuck 😅 

So this is actually merged now, but I have to close it :-)

Tests out fine _but_ I think I might have found another scenario that probably was also broken before: If I join a 2nd player to a multiplayer game, damage that player, log it out, then back in the regen doesn't resume. At that point even re-damaging that player or even respawning doesn't seem to let regen work again.

PR #29, opened 2019-07-07
Update sorted map for base regen case

e-aakash:
Currently, server crashes with NPE on the second player leaving the game.


src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -122,6 +127,8 @@ private void regenerate(long currentTime) {
             RegenComponent regen = entity.getComponent(RegenComponent.class);
             HealthComponent health = entity.getComponent(HealthComponent.class);
             if (health != null && health.nextRegenTick < currentTime) {
+                logger.warn("regenerate reached " + entity.getId() + " value " + regen.getRegenValue() + " rate " +
```
Cervator, line 130: Trivial: Don't need to two-line this, the max line length as per our Checkstyle config is 175, this fits easily (although default IntelliJ  is shorter and might have shown this go past it)

Cervator (approved):
Can't provoke the crash with this merged and confirmed good regen in both headless and headed multiplayer. Merging 👍

PR #36, opened 2020-03-03
Height extras

godfather2:
## Contains
An attempt to fix a sub issue of https://github.com/MovingBlocks/Terasology/issues/2535 along with https://github.com/MovingBlocks/Terasology/pull/3841
The vertical damage speed threshold varies with the player's height. I noticed that at  heights greater than 30,the player took damage on horizontal collision.Hence, i made the character immune to horizontal damage at those heights.

- [x] Increase/Decrease damage from fall. (30m giant should not die of a 15m fall). Define function/proportionality constant same as above.
## How to test

Change the player's height and you will see that the it won't lose any health on jumping.

### DIFF
--- a/src/main/java/org/terasology/logic/health/DamageAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/DamageAuthoritySystem.java
@@ -94,6 +94,13 @@ public class DamageAuthoritySystem extends BaseComponentSystem {
         event.consume();
     }
 
+    private float jumpOfCharacter(EntityRef entity){
+        HealthComponent health = entity.getComponent(HealthComponent.class);
+        CharacterMovementComponent charComp = entity.getComponent(CharacterMovementComponent.class);
+        return charComp.getJumpSpeed();
+
+    }
+
     private void doDamage(EntityRef entity, int damageAmount, Prefab damageType, EntityRef instigator, EntityRef directCause) {
         HealthComponent health = entity.getComponent(HealthComponent.class);
         CharacterMovementComponent characterMovementComponent = entity.getComponent(CharacterMovementComponent.class);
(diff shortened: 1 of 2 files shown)


src/main/java/org/terasology/logic/health/HealthComponent.java
```diff
@@ -16,6 +16,11 @@
 package org.terasology.logic.health;
 
 import org.terasology.entitySystem.Component;
+import org.terasology.entitySystem.entity.EntityRef;
+import org.terasology.logic.characters.CharacterMovementComponent;
+import org.terasology.logic.characters.GazeMountPointComponent;
+import org.terasology.math.geom.Vector3f;
+import org.terasology.network.ClientComponent;
```
e-aakash: These are unused imports.

src/main/java/org/terasology/logic/health/DamageAuthoritySystem.java
```diff
@@ -198,8 +208,9 @@ public void onCrash(HorizontalCollisionEvent event, EntityRef entity, HealthComp
         Vector3f vel = new Vector3f(event.getVelocity());
         vel.y = 0;
         float speed = vel.length();
+        float height = heightOfCharacter(entity);
 
-        highSpeedDamage(speed, entity, health.horizontalDamageSpeedThreshold, health.excessSpeedDamageMultiplier);
+        highSpeedDamage(speed, entity, height < 30 ? health.horizontalDamageSpeedThreshold : 10000, health.excessSpeedDamageMultiplier);
```
e-aakash: Should this be hardcoded? maybe this can be moved to some constant in health component.

src/main/java/org/terasology/logic/health/HealthComponent.java
```diff
@@ -29,6 +34,9 @@
     @Replicate
     public int maxHealth = 20;
 
+
```
e-aakash: Unnecessary newlines added?

e-aakash (commented):
Haven't tested yet, but these are some things that I noticed. 
Maybe values can be extracted away to HealthComponent (making it customizable).

skaldarnar (changes requested):
As stated in this comment https://github.com/MovingBlocks/Terasology/pull/3841#issuecomment-616105139 the health system should react on a scale event, but it should not modify the components `onCrash` or `onLanding`... 

Thus, requesting changes on this PR. I'll create the corresponding event in engine and use it there to adjust values of CharacterComponent and CharacterMovementComponent. You can take that as reference for adjusting the health sytem then 😉

PR #38, opened 2020-04-09
feat: move logic from regen component to system

### LINKED ISSUE #20: Make things more ECS way
Currently, the `RegenComponent` has many logic bits in it. ECS asks for components with only data, and logic should be kept away for the system to handle.

vedant-shroff:
### Contains

Fixed #20 to make the RegenComponent fit better with ECS by removing logic in the RegenComponent and moving it to the RegenAuthoritySystem.

### How to test

Take some damage via fall damage or using the 'damage <amount>' command in the console and wait for your health to regenerate!


skaldarnar (commented):
The build fails due to the test not building anymore. There are tests to ensure a specific regen rate given that two potions are applied. 
I'm wondering how the methods that reside on the system now can be tested...

PR #39, opened 2020-04-18
fix: Remove redundant assets and use CoreAssets

skaldarnar:


jdrueckert (approved):
Tried it out, heart textures and block damage effects look as expected

PR #40, opened 2020-04-30
tweaked logic for joml vectors for particles

pollend:


Cervator:
Redirected PR to develop and merged. Did a version patch bump as well, could actually also push to master now (will fail if not version unique as release management is enabled here)

PR #45, opened 2020-06-19
Health HUD: change maximum icons when maxHealth of entity is changed.

ktksan:
This PR changes the maximum number of health icons that are displayed whenever the maxHealth of an entity is changed.


ktksan:
I think we can place the changeMaxHealth method (currently in ChangeMaxHealthAuthoritySystem) in another system.. but I'm not sure which one (if at all)...

src/main/java/org/terasology/logic/health/event/ChangeMaxHealthEvent.java
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.logic.health.event;
+
+import org.terasology.entitySystem.event.Event;
+
+/**
+ * This Event is sent out whenever a system wants to alter the maxHealth of an entity.
+ */
+public class ChangeMaxHealthEvent implements Event {
```
jdrueckert: IMO it makes sense if other entities are able to "join the discussion", so consider making this an `AbstractValueModifiableEvent` and adjust the Health-internal receiver to be the last entity reacting to this event (see comments above)

src/main/java/org/terasology/logic/health/ChangeMaxHealthAuthoritySystem.java
```diff
@@ -0,0 +1,28 @@
...
+import org.terasology.rendering.nui.NUIManager;
+import org.terasology.rendering.nui.widgets.UIIconBar;
+
+@RegisterSystem(value = RegisterMode.AUTHORITY)
+public class ChangeMaxHealthAuthoritySystem extends BaseComponentSystem {
+    @In
+    NUIManager nuiManager;
+
+    @ReceiveEvent
+    public void changeMaxHealth(ChangeMaxHealthEvent event, EntityRef player, HealthComponent health) {
```
jdrueckert: this should be the last receiver of the event, so that no other entity can affect the max health after we updated the bar, so we should
- have the lowest priority
- consume the event

src/main/java/org/terasology/logic/health/ChangeMaxHealthAuthoritySystem.java
```diff
@@ -0,0 +1,28 @@
...
+import org.terasology.entitySystem.systems.BaseComponentSystem;
+import org.terasology.entitySystem.systems.RegisterMode;
+import org.terasology.entitySystem.systems.RegisterSystem;
+import org.terasology.logic.health.event.ChangeMaxHealthEvent;
+import org.terasology.registry.In;
+import org.terasology.rendering.nui.NUIManager;
+import org.terasology.rendering.nui.widgets.UIIconBar;
+
+@RegisterSystem(value = RegisterMode.AUTHORITY)
+public class ChangeMaxHealthAuthoritySystem extends BaseComponentSystem {
```
jdrueckert: ideally, other entities should be notified when such a change took place, so we should send out a notification event (immutable) after changing max health - we can further consider whether or not to let the ui update react to that instead of the change event

jdrueckert (changes requested):

skaldarnar:
> I think we can place the `changeMaxHealth` method (currently in `ChangeMaxHealthAuthoritySystem`) in another system.. but I'm not sure which one (if at all)...

I agree - `ChangeMaxHealthAuthoritySystem` is awfully specific for this single event handler. However, `DamageAuthoritySystem` handles everything related to damage, and `RestorationAuthoritySystem` handles the opposite part everything related to healing.
We now could split up the event for modify the max health into "damage" and "restoration" but I think that would be over-engineering.

Instead, you could just simplify the name to `HealthAuthoritySystem` for general health-related event handlers. Imho this would only be responsible for setting the value(s) on the component and sent out notifications events (caught by the `HealthClientSystem` to update the UI, for instance).

src/main/java/org/terasology/logic/health/ChangeMaxHealthAuthoritySystem.java
ktksan: AbstractValueModifiableEvent is not consumable though...  But, Changed the priority.

src/main/java/org/terasology/logic/health/event/ChangeMaxHealthEvent.java
```diff
@@ -0,0 +1,15 @@
...
+import org.terasology.entitySystem.event.AbstractValueModifiableEvent;
+
+/**
+ * This Event is sent out whenever a system wants to alter the maxHealth of an entity.
+ */
+public class ChangeMaxHealthEvent extends AbstractValueModifiableEvent {
+    public ChangeMaxHealthEvent(float baseValue) {
+        super(baseValue);
+    }
+}
```
jdrueckert:
```suggestion
}

```

jdrueckert (approved):

PR #47, opened 2020-07-02
feat: Adjust currentHealth on MaxHealthChanged

ktksan:
The PR reduces the currentHealth of the entity to the newMaxHealth in case the currentHealth is greater than newMaxHealth.


src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -20,13 +20,21 @@
...
     NUIManager nuiManager;
 
+    /**
+     * Sends out an immutable notification event when maxHealth of a character is changed.
+     */
     @ReceiveEvent(priority = EventPriority.PRIORITY_TRIVIAL)
     public void changeMaxHealth(ChangeMaxHealthEvent event, EntityRef player, HealthComponent health) {
         player.send(new OnMaxHealthChangedEvent(health.maxHealth, (int) event.getResultValue()));
         health.maxHealth = (int) event.getResultValue();
+        health.currentHealth = Math.min(health.currentHealth, health.maxHealth);
```
jdrueckert: 1. please first change the values and send out the notification(s) afterwards
2. if current health is changed, we need to send out a notification for that too

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -20,13 +20,21 @@
...
         player.saveComponent(health);
     }
 
+    /**
+     * Reacts to the {@link OnMaxHealthChangedEvent} notification event.
+     * Is responsible for the change in maximum number of icons in the Health Bar UI.
+     */
     @ReceiveEvent
     public void onMaxHealthChanged(OnMaxHealthChangedEvent event, EntityRef player) {
```
jdrueckert: Also just realized (must've missed that before), we have a `HealthChanged` event in `Health`, so maybe the corresponding event for max health should be `MaxHealthChanged` instead of `OnMaxHealthChanged`...

src/main/java/org/terasology/logic/health/BlockDamageAuthoritySystem.java
```diff
@@ -197,10 +197,10 @@ private void createBlockParticleEffect(BlockFamily family, Vector3f location) {
         final Stream<Vector2f> baseOffsets = Arrays.stream(BlockPart.sideValues()).map(blockAppearance::getTextureAtlasPos);
 
         return baseOffsets.flatMap(baseOffset ->
-                    IntStream.range(0, 8).boxed().map(i ->
+                IntStream.range(0, 8).boxed().map(i ->
```
ktksan: I tried using git reset -- HEAD command for this file 
but for some reason, this still shows up here.......😅

ktksan: Never mind, done. I realized I was comparing it to my fork...

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
```diff
@@ -19,16 +22,29 @@
...
+    /**
+     * Sends out an immutable notification event when maxHealth of a character is changed.
+     */
     @ReceiveEvent(priority = EventPriority.PRIORITY_TRIVIAL)
     public void changeMaxHealth(ChangeMaxHealthEvent event, EntityRef player, HealthComponent health) {
-        player.send(new OnMaxHealthChangedEvent(health.maxHealth, (int) event.getResultValue()));
+        int oldMaxHealth = health.maxHealth;
         health.maxHealth = (int) event.getResultValue();
+        Prefab maxHealthReductionDamagePrefab = prefabManager.getPrefab("Health:maxHealthReductionDamage");
+        player.send(new DoDamageEvent(Math.max(health.currentHealth - health.maxHealth, 0),
```
jdrueckert, line 36: 1. Do we really want to use a `DoDamageEvent` here? The health change is not really damage-inflicted, is it? :thinking: 
2. Tbh I'm not entirely sure whether this should be in `changeMaxHealth` or rather in `onMaxHealthChanged`...

ktksan: 1. Yeah, I think we are doing damage, since maxHealth is being reduced damage is being dealt, and well the DoDamageEvent does have a prefab which stores information regarding the type of damage being dealt. And hence the maxHealthReductionDamagePrefab is enough to convey a damage is being dealt which is because we are reducing the player's maxHealth and the currentHealth is greater than the new maxHealth. Another thought, reducingMaxHealth is in itself a damage in a way just a little more permanent than a regular one. 

2. I think CurrentHealth should not be greater than maxHealth at any point of time another event handler can react to, thus the change should happen before a notification for maxHealth change is sent. So that if any of the event handlers reacts to the notification, then at the time currentHealth does not exceed the maxHealth even there.

jdrueckert: wrt (2) okay fair point.
wrt (1) I disagree in that we're doing damage. IMO not all health decreases are inflicted by damage - the English language knows many subtle differences here. From how I interpret it, damage is something that is inflicted from outside, either on purpose or by accident. Here however, the max health is impacted resulting in the actual health going down as well. Realistic examples might be becoming weaker with age or due to an illness - that's a difference to getting hit by somebody or something and suffering damage as a result.

ktksan: Actually, I do not disagree with you in terms of whether it is actually a "damage"... I mean I was leaning towards not calling it damage in the first place which is why I didn't use DoDamageEvent in my first iteration of the method.

But, only reason I think doDamageEvent should be used is because when currentHealth is changed, a HealthChangedEvent needs to be sent out. Now, this event is never used directly, 2 other events - OnDamagedEvent, OnRestored event implement this event. So I had to use either of the 2, among which OnDamagedEvent is clearly a better choice. Now, since we are already sending an OnDamagedEvent, I figured I might as well send a prefab specifying the type of damage which is a parameter of DoDamageEvent (actually now I see, OnDamagedEvent also has the damageType parameter).... 
SideNote: I could've used HealthChangedEvent directly but the event does not specify whether the change in health was incremental or decremental.
So.. my options are:
1. leave it as is.
2. Use HealthChangedEvent directly (which doesn't specify the type of change)
3. Use OnDamagedEvent directly (which would also be more or less same as the current situation only difference - the currentHealth will be reduced here and not in the DamageAuthoritySystem)..

skaldarnar: Thanks for the explanation! 

Nitpick question: What happens if we send a DoDamageEvent with amount zero? Is that automatically canceled, or does that result in some audio cue of being hurt without actually being hurt?

skaldarnar (approved):

src/main/java/org/terasology/logic/health/HealthAuthoritySystem.java
ktksan: Yep 0 damage us ignored https://github.com/Terasology/Health/blob/178462a516e6b58d60bb2355709de916262236da/src/main/java/org/terasology/logic/health/DamageAuthoritySystem.java#L128-L129

skaldarnar:
Failing tests are tracked in #49 - merging this despite those two failing tests now 🙃

PR #48, opened 2020-07-12
doc: Improve JavaDoc for ActivateRegenEvent

skaldarnar:
Improve JavaDoc for `ActiveRegenEvent` by making some internal assumptions and semantics explicit.

For instance, the `endTime`  of the activate event is _not_ a time stamp when the effect should end (in world time) but is interpreted as the **duration** the effect should stay active in seconds. I think it is important to state the unit (seconds) that the value is assumed to be in, as well as  what it is actually used for. 

Same holds for the `id` and the `value`.

### DIFF
--- a/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
@@ -1,26 +1,11 @@
-/*
- * Copyright 2019 MovingBlocks
- *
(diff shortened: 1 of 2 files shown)


e-aakash (approved):
A very detailed documentation addition. 
This can act as a guide on how to write docs for beginners (+ me 🙂  )

skaldarnar:
I'm going to merge this doc PR despite the failing tests as they were not introduced by this PR, and solution for the flaky tests still pending (see #49).

PR #49, opened 2020-07-12
test: fix failing tests by increasing timeout

skaldarnar:
I noticed in #48 that some MTE tests are failing. This seems be to caused by some timeouts being to low (0.2 seconds for extra processing). This PR increases the extra time to 0.5 seconds - tests are succeeding locally for me.

### DIFF
--- a/assets/prefabs/damageTypes/maxHealthReductionDamage.prefab
+++ b/assets/prefabs/damageTypes/maxHealthReductionDamage.prefab
@@ -1,0 +1,5 @@
+{
+    "DisplayName": {
+        "name": "MaxHealthReduction"
+    }
+}
--- a/module.txt
+++ b/module.txt
@@ -1,6 +1,6 @@
 {
     "id" : "Health",
-    "version" : "1.0.1",
+    "version" : "1.0.2",
     "isReleaseManaged": true,
     "author" : "The Terasology Foundation",
     "displayName" : "Health",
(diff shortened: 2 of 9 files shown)


skaldarnar:
Obviously, this didn't help :thinking:

Cervator:
Yeah it sounds like one test was still just barely off point (95 vs 100 - one "tick") ? Timing-sensitive tests suck 😞 

Not sure why CI seems more sensitive. Can't say with the amount of logging if it has anything to do with Jenkins itself. You've got a GitHub Action that works with unit tests - might it be worth trying that setup to see if it works there? Although the main difference there may be full engine workspace + module dropped in vs testing in a standalone module workspace. It might be worth trying the test standalone locally, but I'm skeptical if that'll differ.

skaldarnar:
The failing tests seem to be affected by other tests that run before - albeit theoretically being separated :thinking: Maybe @kaen stumbles upon something while working on the MTE. Could using the JUnit 5 extension help here?

Cervator:
When in doubt with quirky issues: do a major version upgrade and hope for the best :-)

That _could_ possibly help, if the way stuff is being prepared changes. At least we should try that first since we should do that upgrade anyway, probably better to try that first than attempt quirky fixes on a doomed older version.

skaldarnar:
I was about to say "this branch is open for contribution" but then I realized it's from a fork - automagically created by `gh` :upside_down_face: I'll try to update the MTE tests to JUnit 5 extension within the next days, see whether that helps with the flaky tests...

PR #51, opened 2020-11-17
fix: regeneration starting too soon

skaldarnar:
The regeneration logic for blocks was broken such that the regeneration would start already while the block was still actively damaged.
These changes will put a delay via the DelayManager for activating the regeneration.

This introduces a breaking change to ActivateRegenEvent which now requires the regen value to be specified also for the base regen.

### DIFF
--- a/src/main/java/org/terasology/logic/health/DamageAuthoritySystem.java
+++ b/src/main/java/org/terasology/logic/health/DamageAuthoritySystem.java
@@ -1,20 +1,9 @@
-/*
- * Copyright 2019 MovingBlocks
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// Copyright 2020 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
 package org.terasology.logic.health;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.terasology.audio.StaticSound;
 import org.terasology.audio.events.PlaySoundEvent;
 import org.terasology.audio.events.PlaySoundForOwnerEvent;
@@ -32,9 +21,12 @@ import org.terasology.logic.characters.MovementMode;
 import org.terasology.logic.characters.events.AttackEvent;
 import org.terasology.logic.characters.events.HorizontalCollisionEvent;
 import org.terasology.logic.characters.events.VerticalCollisionEvent;
+import org.terasology.logic.delay.DelayManager;
+import org.terasology.logic.delay.DelayedActionTriggeredEvent;
 import org.terasology.logic.health.event.ActivateRegenEvent;
 import org.terasology.logic.health.event.BeforeDamagedEvent;
 import org.terasology.logic.health.event.DamageSoundComponent;
+import org.terasology.logic.health.event.DeactivateRegenEvent;
 import org.terasology.logic.health.event.DoDamageEvent;
 import org.terasology.logic.health.event.DoRestoreEvent;
 import org.terasology.logic.health.event.OnDamagedEvent;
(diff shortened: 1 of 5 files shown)


jdrueckert:
Requires to be merged together with https://github.com/Terasology/ClimateConditions/pull/35.
Otherwise this breaks omega compilation.

src/main/java/org/terasology/logic/health/DamageAuthoritySystem.java
```diff
@@ -103,32 +106,55 @@ private void doDamage(EntityRef entity, int damageAmount, Prefab damageType, Ent
...
+        }
+        long delayInMs = Math.round(delayInSeconds * 1000);
+        logger.debug("Scheduling delayed regen event with delay '{}'", delayInMs);
+        delayManager.addDelayedAction(entity, DELAYED_REGEN_ACTIVATION, delayInMs);
+    }
+
+    @ReceiveEvent
+    public void onDelayedRegenActivation(DelayedActionTriggeredEvent event, EntityRef entity, HealthComponent health) {
+        if (event.getActionId().equals(DELAYED_REGEN_ACTIVATION)) {
+            logger.debug("Sending delayed regen event");
```
jdrueckert:
```suggestion
            logger.debug("Sending regen event");
```
or
```suggestion
            logger.debug("Sending regen event after delay has passed");
```

because it's no longer delayed now.

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -35,17 +37,19 @@
...
-    public static final String WAIT = "wait";
 
     /**
      * Integer storing when to check each effect.
      */
     private static final int CHECK_INTERVAL = 200;
 
     /**
-     * Long storing when entities are to be regenerated again.
+     * The in-game in ms at which entities are to be regenerated again.
```
jdrueckert:
```suggestion
     * The in-game time in ms at which entities are to be regenerated again.
```
?

src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -24,16 +11,35 @@
 import java.util.HashMap;
 import java.util.Map;
 
-import static org.terasology.logic.health.RegenAuthoritySystem.BASE_REGEN;
-
+/**
+ * Not for direct access! Use regen events instead.
+ *
+ * @see org.terasology.logic.health.event.ActivateRegenEvent
+ * @see org.terasology.logic.health.event.DeactivateRegenEvent
```
jdrueckert, line 18: link? or is `@see` like link?

src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -24,16 +11,35 @@
...
+ */
 public class RegenComponent implements Component {
+    /**
+     * The timestamp in in-game time (ms) when the next regeneration action is due (or is to end?).
```
jdrueckert: > (or is to end?)

should this stay here?

src/main/java/org/terasology/logic/health/RegenComponent.java
```diff
@@ -24,16 +11,35 @@
...
     public long soonestEndTime = Long.MAX_VALUE;
+
+    /**
+     * Mapping from regeneration action ids to the regeneration value.
+     */
     @Replicate
     public Map<String, Float> regenValue = new HashMap<>();
+
+    /**
+     * Registered regeneration action ids associated to their end time(?).
```
jdrueckert: > (?)

deliberate?

jdrueckert (changes requested):

src/main/java/org/terasology/logic/health/RegenComponent.java
skaldarnar: I have no clue :shrug: nobody uses this anyways... :roll_eyes:

src/main/java/org/terasology/logic/health/RegenComponent.java
skaldarnar:
```suggestion
     * Registered regeneration action ids associated to their end time.
```

jdrueckert (changes requested):
The latest commit seems to have broken something... Regeneration is no longer working.

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -186,18 +185,20 @@ public void onRegenAddedWithoutComponent(ActivateRegenEvent event, EntityRef ent
...
-                                     HealthComponent health) {
-        if (event.id.equals(BASE_REGEN)) {
-            // setting endTime to -1 because natural regen happens till entity fully regenerates
-            addRegen(BASE_REGEN, health.regenRate, -1, regen);
-        } else {
-            addRegen(event.id, event.value, time.getGameTimeInMs() + (long) (event.endTime * 1000), regen);
+    private void addRegenToScheduler(ActivateRegenEvent event, RegenComponent regen) {
+        if (event.value != 0) {
+            // handle indefinite regeneration actions
+            final long endTime = event.value < 0 ? -1 : time.getGameTimeInMs() + (long) (event.endTime * 1000);
```
skaldarnar: @jdrueckert I'm pretty sure this is the culprit:

```suggestion
            final long endTime = event.endTime < 0 ? -1 : time.getGameTimeInMs() + (long) (event.endTime * 1000);
```

jdrueckert (approved):

jdrueckert:
Failing tests were failing before, too.

PR #53, opened 2020-11-23
feat: Add a red flash on receiving damage

askneller:
### What this PR does

Adds a red flash on the player receiving damage (the death screen background). Overlay stays for 150ms.

image

### How to Test

Start a game with WildAnimals enabled and spawn in a hostileDeer. You should see the flash when it attacks you.

Requires https://github.com/MovingBlocks/TeraNUI/pull/42


skaldarnar:
Hej @askneller , thanks a lot for the PR! 🚀 

Some comments/questions even before I do a proper review:

- I think we already have a slight camera shake on hit (but I might be mistaken). Thus, I'm wondering whether we can limit this more drastic feedback to "hard hits", e.g., when the player takes more than X damage in one hit, or looses more than Y% of their health with a single hit. What do you think of that? 
- Many games these days use directional indicators to give a hint to the player where they are damaged from. How complicated would that be here? I could easily provide some assets for this if we want to try it out...

jdrueckert:
> or looses more than Y% of their health with a single hit

Maybe Y% of their _current_ health... Thus it will get more frequent and visible the less health the player gets. If you only have 2 hearts left, losing half a heart may already be a hard hit.

askneller:
@skaldarnar @jdrueckert  Thanks for the comments. Skaldarnar I can give the directional indicator a try, some assets would be nice thanks.

Jdrueckert how about 25% of current health in one hit?

jdrueckert:
> how about 25% of current health in one hit?

Yes I think that sounds like a reasonable value. If during playtests we see that it's too often or maybe not often enough, we can still change that value in the future to improve balancing if necessary.

skaldarnar:
@askneller @jdrueckert Here are two variants, one quite simple spanning the whole side, the other slightly triangle shaped only in the center. I can easily export each layer as separate image to have the individual sites separate. Probably a single texture rendered multiple times in the different locations is sufficient, though.

Let me know which one you prefer.

### Line
terasology_hit-indicator-simple
hit-indicator-line

### Triangle
terasology_hit-indicator-triangle
hit-indicator-triangle

askneller:
I have updated the PR so that the flash only appears when the player receives damage >= 25% of their current health. I was also able to add the directional indicator. @skaldarnar I used the triangle one but unfortunately I think the transparency was lost when you pasted it (it has a white background).

Note for the direction to work this requires this PR: https://github.com/Terasology/Behaviors/pull/54

skaldarnar:
Thanks a lot @askneller - will check it out today!

src/main/java/org/terasology/logic/health/HealthClientSystem.java
```diff
@@ -57,51 +62,77 @@ public void initialise() {
...
             }
         }
     }
 
-    private double rad2Deg(double rad) {
-        return rad / Math.PI * 180.0;
+    private double determineDamageDirection(EntityRef instigator, LocationComponent locationComponent) {
+        LocationComponent instigatorLocation = instigator.getComponent(LocationComponent.class);
+        Vector3f loc = new Vector3f();
+        loc = locationComponent.getWorldPosition(loc);
```
skaldarnar: These two lines create a new vector `loc` and then write the world position from the component into that vector - so far, so good. 👍

src/main/java/org/terasology/logic/health/HealthClientSystem.java
```diff
@@ -57,51 +62,77 @@ public void initialise() {
...
+        Vector3f instLoc = new Vector3f();
+        instLoc = instigatorLocation.getWorldPosition(instLoc);
+        Vector3f locDiff = instLoc.sub(loc);
```
skaldarnar: Now this line is a bit "tricky" (as in, I think it works in this context, but bears a pitfall I'd like to explain).

`instLoc.sub(loc)` does the following under the hood:

- it mutates the vector you call the method on (`instLoc`) such that it holds the result afterwards
- it returns that vector after mutation (again, `instLoc`) so that you can chain multiple operations

This means that this line **modifies `instLoc`** and also creates a new _alias_ for it under the name `locDiff`.

src/main/java/org/terasology/logic/health/HealthClientSystem.java
```diff
@@ -57,51 +62,77 @@ public void initialise() {
...
+        locDiff.normalize();
```
skaldarnar: That said, these four lines could be combined to save one allocation of a new vector. 

```suggestion
		Vector3f locDiff = instigatorLocation.getWorldPosition(new Vector3f()).sub(loc).normalize();
```

src/main/java/org/terasology/logic/health/HealthClientSystem.java
```diff
@@ -57,51 +62,77 @@ public void initialise() {
...
+
+        Vector3f worldFacing = new Vector3f();
+        // facing x and z are "how much" of that direction we are facing
+        // e.g. (0.0, 1.0) means that going forward increases world z position without increasing x position
+        worldFacing = locationComponent.getWorldDirection(worldFacing);
+        worldFacing.normalize();
```
skaldarnar: This is likely a matter of taste - you _could_ combine this into a single line:

```suggestion
        // facing x and z are "how much" of that direction we are facing
        // e.g. (0.0, 1.0) means that going forward increases world z position without increasing x position
		Vector3f worldFacing = locationComponent.getWorldDirection(new Vector3f()).normalize();
```

skaldarnar (changes requested):

src/main/java/org/terasology/logic/health/HealthClientSystem.java
askneller: Ah, thanks. I guess I was a bit confused regarding using JOML. I'll keep that in mind.

skaldarnar:
So, testing this in multiplayer revealed that it's not properly working. I've tried to fix it, but this became a bigger issue than expected:

- `OnDamagedEvent` and `OnRestoreEvent` are server-side-only events and never reach the clients - so no way to react on them
- even after adding a default constructor and using `@OwnerEvent` I cannot get the event to be replicated to the clients =/
- in addition, there is this missing type handler (probably not directly related, though)
  ```
  No type handler for type interface com.google.common.collect.SortedSetMultimap used by class org.terasology.logic.health.RegenComponent::regenEndTime
  ```
- the client wants to use the `DelayManager` to perform a delayed action - it seems that the `DelayManager` is not initialized on the client, so the client cannot use it

I don't know how to proceed here, seems to be an impossible task (at least if you want to solve it in the "right" way using the ECS...)

src/main/java/org/terasology/logic/health/HealthClientSystem.java
```diff
@@ -1,34 +1,71 @@
...
+                "directionalDamageOverlay");
+    }
+
+    @ReceiveEvent(components = PlayerCharacterComponent.class)
+    public void onDamaged(OnDamagedEvent event, EntityRef entity,
+                          LocationComponent locationComponent,
+                          HealthComponent healthComponent) {
+        // Show the relevant direction element
+        EntityRef instigator = event.getInstigator();
+        if (instigator != null && instigator.hasComponent(LocationComponent.class)) {
```
jdrueckert, line 43: We also need to look at the cases in which `instigator` is null, e.g. for falling damage or damage due to Hunger and Thirst.
However, it's fine to do this as a follow-up PR and keep the scope of this PR to "add red flash on receiving damage from other entities"

src/main/java/org/terasology/rendering/nui/layers/hud/DirectionalDamageOverlay.java
```diff
@@ -0,0 +1,81 @@
...
+            case FORWARD:
+                return Optional.of(top);
+            case RIGHT:
+                return Optional.of(right);
+            case LEFT:
+                return Optional.of(left);
+            case BACKWARD:
+                return Optional.of(bottom);
+            case UP:
+            case DOWN:
```
jdrueckert, line 48: Should these cases return top / bottom? :thinking:

src/test/java/org/terasology/logic/health/HealthClientSystemTest.java
```diff
@@ -0,0 +1,51 @@
...
+    void determineDamageDirectionForCompassDirections(Direction direction) {
+
+        EntityRef instigator = new PojoEntityManager().create();
+        instigator.addComponent(new LocationComponent(direction.asVector3f()));
+
+        Direction damageDirection = healthClientSystem.determineDamageDirection(instigator, location);
+
+        assertEquals(direction, damageDirection);
+    }
+}
```
jdrueckert: newline

src/main/java/org/terasology/rendering/nui/layers/hud/DirectionalDamageOverlay.java
skaldarnar: I couldn't wrap my head around how to best render the indicators if one indicator is used for multiple directions... =/ I think the data structures as they are don't fit that use case.

The concrete problem I had (why the bottom indicator did not show up) was that for `FORWARD`/`UP` the order was "correct", and I first set the image visibility to `false` and then to `true`. For `BACKWARD`/`DOWN` it was the other way around, the second direction canceled out the visibility..

Maybe it would be better to iterate over the images and do the lookup the other way around, and then add some _one-of_ logic to determine whether the indicator is visible...

jdrueckert: Can you write up a follow-up issue for that? Same for the damage indicator for falling and thirst/hunger damage.

skaldarnar: I'd rather directly do the follow-up PR...

skaldarnar (dismissed):

skaldarnar (approved):

jdrueckert (approved):

PR #59, opened 2021-01-09
chore(JOML): align with MTE migration

jdrueckert:
https://github.com/Terasology/ModuleTestingEnvironment/pull/34

### DIFF
--- a/src/test/java/org/terasology/logic/health/BlockTest.java
+++ b/src/test/java/org/terasology/logic/health/BlockTest.java
@@ -33,7 +33,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 public class BlockTest extends ModuleTestingEnvironment {
-    private static final Vector3i BLOCK_LOCATION = new Vector3i().add(0, -1, 0);
+    private static final Vector3ic BLOCK_LOCATION = new Vector3i(0, 0, 0).add(0, -1, 0);
 
     private WorldProvider worldProvider;
     private Time time;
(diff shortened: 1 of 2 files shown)


src/test/java/org/terasology/logic/health/BlockTest.java
```diff
@@ -33,7 +33,7 @@
 import static org.junit.Assert.assertTrue;
 
 public class BlockTest extends ModuleTestingEnvironment {
-    private static final Vector3i BLOCK_LOCATION = Vector3i.zero().add(Vector3i.down());
+    private static final Vector3i BLOCK_LOCATION = new Vector3i().add(0, -1, 0);
```
skaldarnar: As @keturn pointed out before, the `Vector3i.zero()` was explicit in what the vector is, whereas `new Vector3i()` only implicitly holds that information, and might not be clear from looking at the test case. 

```suggestion
    private static final Vector3ic BLOCK_LOCATION = new Vector3i(0, 0, 0).add(0, -1, 0);
```

skaldarnar (changes requested):

src/test/java/org/terasology/logic/health/BlockTest.java
pollend: I also suggest using Vector3ic.

skaldarnar: incorporated Michael's comment in the suggestion 😉

skaldarnar (dismissed):

skaldarnar (approved):

jdrueckert:
Failing tests were failing before.

PR #62, opened 2021-02-05
test: add unit test for 'BlockDamageRenderer#getDamageEffectsNumber'

skaldarnar:


src/test/java/org/terasology/logic/health/BlockDamageRendererTest.java
```diff
@@ -0,0 +1,64 @@
...
+        return Stream.of(
+                Arguments.of(healthComponent(-20, 100)),
+                Arguments.of(healthComponent(-20, -100)),
+                Arguments.of(healthComponent(20, -100))
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("validHealthComponents")
+    void getDamageEffectsNumber_validHealthComponents(int expected, HealthComponent health) {
```
skaldarnar: This follows the naming scheme as defined in [Google's Java Style Guide](https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names).

src/main/java/org/terasology/logic/health/BlockDamageRenderer.java
```diff
@@ -80,8 +83,20 @@ public void renderOverlay() {
...
+    /**
+     * Compute the damage effect number as linear mapping from damage percentage to the range [0..10].
+     * <p>
+     * A value of 0 denotes no damage, while 10 denotes a current health of 0.
+     *
+     * @return the effect number in [0..10] linear to damage percentage
+     */
+    int getDamageEffectsNumber(HealthComponent health) {
+        Preconditions.checkArgument(health.currentHealth >= 0);
+        Preconditions.checkArgument(health.maxHealth > 0);
```
skaldarnar, line 96: I thought a hard crash would make sense in these cases, instead of weird effect numbers... :shrug: But we can discuss this, I'm fine with changing.

src/test/java/org/terasology/logic/health/BlockDamageRendererTest.java
pollend: is there a way to add this to the linting to it doesn't get marked.

pollend (dismissed):

src/test/java/org/terasology/logic/health/BlockDamageRendererTest.java
skaldarnar: probably, but in the meantime I found a way to make better use of `@DisplayName` and would not use this format anymore... will change it and merge then :nerd_face:

pollend (approved):
we can fix the test afterwards.

PR #64, opened 2021-02-11
feat(test): migrate test cases to MTTExtension

pollend:
some of these test were broken in previous commits. any clue what was changed to break this and do we need to modify the module in any way @skaldarnar.


keturn (approved):
LGTM as far as ModuleTestingEnvironment -> MTEExtension stuff goes

PR #65, opened 2021-02-13
fix: avoid NPE in RegenAuthoritySystem

skaldarnar:


skaldarnar:
The tests were failing before, and this incident is not covered by any test.

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
```diff
@@ -122,6 +122,10 @@ private void regenerate(long currentTime) {
         for (EntityRef entity : regenSortedByTime.values()) {
             RegenComponent regen = entity.getComponent(RegenComponent.class);
             HealthComponent health = entity.getComponent(HealthComponent.class);
+            if (regen == null || health == null) {
+                logger.debug("Entity '{}' without RegenComponent or HealthComponent scheduled for regeneration - skipping", entity);
+                continue;
+            }
             if (health != null && health.nextRegenTick < currentTime) {
```
pollend: you don't need this null check after this.

pollend (changes requested):

src/main/java/org/terasology/logic/health/RegenAuthoritySystem.java
skaldarnar:
```suggestion
            if (health.nextRegenTick < currentTime) {
```

pollend (approved):

skaldarnar:
Test failed before.

PR #66, opened 2021-02-21
test: prevent test event handlers from interfering with each other

keturn:
This doesn't fix _all_ the tests, but it does help.


keturn:
I admit the try-blocks are ugly here, as much as is crammed in to that expression creating the resource. I bet there's some way we could do some kind of annotation thing that would inject an event handler and unregister it at the end of the test instead of the try-with-resources blocks. Sounds like something @DarkWeird would know?

DarkWeird:
You can do it via creating ComponentSystem with registering and unregistering at EventSystem.

pollend (approved):

PR #68, opened 2021-02-28
test: fix (some) failing MTE tests

skaldarnar:
This reduces the number of failing tests from 4 to 2. I think we can (and should) merge this as one step in the right direction...


skaldarnar:
Interestingly, this test fails, because the entity keeps regenerating health...

```java
    @Test
    public void regenCancelTest() {
        EntityRef player = newPlayer(100, 1);

        // Deactivate all regen
        player.send(new DeactivateRegenEvent(RegenAuthoritySystem.ALL_REGEN));

        assertFalse(player.hasComponent(RegenComponent.class));
        int currentHealth = player.getComponent(HealthComponent.class).currentHealth;
        assertTrue(currentHealth < 100, "regeneration should have been canceled before reaching max health");

        float tick = time.getGameTime() + 5f;
        helper.runWhile(()-> time.getGameTime() <= tick);

		 // FAILS HERE!
        assertEquals(currentHealth, player.getComponent(HealthComponent.class).currentHealth); 
    }
```

Maybe the test itself should end when checking that some specific component is there, i.e., have each test only check a single thing...

On the other hand, this may actually point to a bug here, as regeneration continues even though the `RegenComponent` is not present anymore.

keturn (approved):
I think we should be very picky about how many of these time-based tests we include in the test suite.

But I do like passing tests! As well as removing duplicate tests. I'll accept this as improvement over the current state of things.

src/test/java/org/terasology/logic/health/RegenTest.java
```diff
@@ -99,4 +106,32 @@ public void zeroRegenTest() {
...
+        healthComponent.maxHealth = 100;
+        healthComponent.waitBeforeRegen = 1;
+        healthComponent.regenRate = 1;
+
+        final EntityRef player = entityManager.create();
+        player.addComponent(new PlayerCharacterComponent());
+        player.addComponent(healthComponent);
+
+        player.send(new DoDamageEvent(5));
+        helper.runWhile(200, () -> true);
```
keturn, line 123: Given a no-op, could this be Thread.sleep?

src/test/java/org/terasology/logic/health/RegenTest.java
```diff
@@ -99,4 +106,32 @@ public void zeroRegenTest() {
...
+        assertEquals(healthComponent.currentHealth, 95);
+
+        // wait 1 second before regen starts
+        float regenStarts = time.getGameTime() + healthComponent.waitBeforeRegen + BUFFER;
+        helper.runWhile(() -> time.getGameTime() <= regenStarts);
```
keturn, line 128: and this is like a `sleep` but in GameTime units. Not quite sure how to do that one.

skaldarnar: I thought the `runWhile`/`runUntil` helper methods are exactly for this purpose? :thinking: I think in other places I uses a different pattern:

```
helper.runWhile(healthComponent.waitBeforeRegen + BUFFER, () -> true)
```

This runs indefinitely, but at max until the timeout of `healthComponent.waitBeforeRegen + BUFFER` is reached (I think the timeout uses a `long` for ms  while the game time here is `float` in seconds).

src/test/java/org/terasology/logic/health/RegenTest.java
skaldarnar: Maybe? :thinking: I don't fully get how the game ticks work here... this is probably not needed, and just an artifact of me tinkering around ... :monocle_face: 

I'll do another round over these tests anyways, will see how much I can throw away then :sweat_smile:

PR #69, opened 2021-02-28
Test/regen event handlers

keturn:


keturn:
oops, this was already merged, just squashed

PR #74, opened 2021-03-18
doc: create gh-pages with docsify

skaldarnar:
The [docsify page](https://docsify.js.org/#/) is previewed under my fork: https://skaldarnar.github.io/Health/#/

- doc: add gh-page based on docsify
- doc: add full text search plugin
- doc: add 'edit on github' plugin
- doc: set theme color and add icon
- doc: update README with banner image and link to documentation page

Along with this PR, I'd like to start some discussion on **module conventions**: I think we should find some convention for modules that we can utilize for the ModuleSite, e.g, 

- have a README with similar format (front matter?)
- have a banner image in the same location, e.g., `docs/banner.png` or `docs/_media/banner.png`
- provide extended information/documentation somewhere, e.g,. their wiki or **github pages** (preferred)

I do believe switching out these simple static site generators is fairly simple task, so exploring this in two or three directions should not hurt that much. (I just need to make sure to bring this topic up often enough so that we decide on one version at some point...)

--- 

_a bit of background information_

> **my pain point:** we have documentation in too many places, and it is no fun to update them. likewise, it is hard to find information for a specific topic. 
> 
> While ModuleSite is actively developed, it does not solve the problem of content creation. It should aggregate module info, but where is that coming from? 
> 
> I'm not happy with how our tutorials are yet another module, and their tutorial text is (technically) living in yet another repo (the wiki). Updating tutorials and keeping them in sync with the current state of the code is already hard, but now I have to open my browser and my IDE to do the changes. 
> 
> In addition, many of our modules don't have documentation at all, although I think it would help having that (e.g., a high-level explanation of what the API of the Health module is). I'm a big fan of having that information close to the code, that it is very easy to just update it alongside. 
> 
> Therefore, I wanted to explore the space of using gh-pages for that purpose, hopefully without the need of any CI steps to get it done.


docs/README.md
```diff
@@ -0,0 +1,43 @@
+<div align="center">
+<img src="./_media/banner.png">
+</div>
+
+_This is a module for [Terasology].
+It adds the concept of health (or hitpoints) to the game and handles regeneration, restoration and damage of entities._
```
keturn: would be nice if we didn't have to duplicate /README.md and /doc/README.md.

that seems like it might be fixable by configuration.

docs/index.html
```diff
@@ -0,0 +1,36 @@
...
+      search: 'auto',
+      plugins: [
+      EditOnGithubPlugin.create(
+          "https://github.com/Terasology/Health/tree/develop/docs/"
+        )
+      ]
+    }
+  </script>
+  <!-- Docsify v4 -->
+  <script src="//cdn.jsdelivr.net/npm/docsify@4"></script>
```
keturn, line 34: Oh, you pull the whole docsify thing out of CDN. Its themes too, I see!

That's nice. I was going to worry if every module had to include it in its repo.

docs/regeneration.md
```diff
@@ -0,0 +1,9 @@
+# Regeneration
+
+Handles the natural healing of entities (and blocks).
+To activate regeneration send `ActivateRegenEvent(String id, float value, float endTime)`. 
+Health is regenerated every second. 
+Empty event `ActivateRegenEvent()` activates base regeneration of entity.
+
+To deactivate particular type of regeneration, send `DeactivateRegenEvent(String id)`. 
+Empty event `DeactivateRegenEvent()` deactivates base regeneration fo entity.
```
keturn, line 9: I assume these are skeleton pages and you expect them to have more content?

keturn (commented):
The question I brought up in #documentation is: Since all the content is fetched by javascript, does it still show up in a third-party search engine?

I think to find this out you'd go to, what's it called these days, Google Search Console and use its [URL Inspection](https://search.google.com/search-console/inspect) thing to see what the indexed site looks like.

(Does DuckDuckGo have a similar thing? [Guess not.](https://www.reddit.com/r/duckduckgo/wiki/index#wiki_webmaster_tools))

The other thing I would wonder is:

Not all documentation is API documentation, but I _do_ find API documentation useful as well.

I guess there _could_ be a browser-side JavaDoc compiler. JavaDoc is HTML-ish and everything. But I didn't see it on the list of docsify addons.

That's not a blocker for making these non-API docs available but does inform how to proceed. If each module is going to build its JavaDoc server-side, might as well do the Markdown server-side too.

Or maybe figuring out how to do the cross-linking between a hundred different JavaDoc publications is a nasty problem, so we put up a single `api-docs.terasology.org` populated with omega and individual module sites don't have to worry about it at all!

keturn:
overall, I'm +1 to having docs and source in the same repo, and am satisfied by the low amount of extra _stuff_ that needs to be added to the repo since it's all served out of someone else's CDN.

docs/README.md
skaldarnar: Yeah, I don't like that, but I also don't know how that would work 🤔  
- the top-level README is what github displays if you just go the repo, and I think we want that
- github pages is served out of the `docs/` folder, and I don't think that an "up-reference" `..` is possible (not sure, though)

I did a brief search for that, but did not find anything on first glance...

docs/regeneration.md
skaldarnar: Yes, this is what was present in the README before, but I'd like to expand them to contain more information in the future.

skaldarnar:
@keturn re Javadoc: we manually publish the Javadoc on Github pages for the https://github.com/Terasology/ModuleTestingEnvironment. Someone needs to run `gradlew :modules:ModuleTestingEnvironment:javadoc` (or something like that) manually, copy the output to the `docs` folder and commit it. While this works, this is (a) a huge diff, and (b) quite some noise on the repo, and (c) should ideally be done after each commit to ensure docs are up to date with the code. You can see the result here: https://terasology.github.io/ModuleTestingEnvironment/

Afaik our Jenkins also builds the docs, but does not publish/serve them anywhere. Maybe we can just utilize https://github.com/jenkinsci/javadoc-plugin in some way, and have both the README and the documentation side link to it?

jdrueckert (approved):

PR #79, opened 2021-04-18
refactor(!): gradually improve events

skaldarnar:
I recommend to go over this change set commit by commit.
	
This is technically a breaking change, but nothing in Omega seems to be using the affected API directly, so practically this change is local to the module. No module PRs are required.

### DIFF
--- a/src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
+++ b/src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
@@ -10,7 +10,7 @@ import org.terasology.engine.network.OwnerEvent;
  * This event is sent after damage has been dealt to an entity.
  */
 @OwnerEvent
-public class OnDamagedEvent extends HealthChangedEvent {
+public class OnDamagedEvent extends OnHealthChangedEvent {
     private Prefab damageType;
     private int fullAmount;
 
--- a/src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
+++ b/src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.terasology.module.health.events;
 
+import com.google.common.base.Preconditions;
 import org.terasology.engine.entitySystem.entity.EntityRef;
 import org.terasology.engine.entitySystem.prefab.Prefab;
 import org.terasology.engine.network.OwnerEvent;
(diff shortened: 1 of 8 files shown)


skaldarnar:

src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
```diff
@@ -11,20 +12,16 @@
...
 
-    public OnDamagedEvent() {
-    }
-
-    public OnDamagedEvent(int fullAmount, int change, Prefab damageType, EntityRef instigator) {
-        super(instigator, change);
-        this.fullAmount = fullAmount;
+    public OnDamagedEvent(int change, Prefab damageType, EntityRef instigator) {
+        super(-change, instigator);
+        Preconditions.checkArgument(change >= 0, "damage amount must be non-negative - use OnRestoredEvent instead");
```
jdrueckert, line 19: should this arg check be done before calling `super()`? :thinking:

src/main/java/org/terasology/module/health/events/OnRestoredEvent.java
```diff
@@ -10,21 +11,14 @@
...
-    private int amount;
-
-    public OnRestoredEvent() {
-    }
 
     public OnRestoredEvent(int amount, EntityRef instigator) {
-        super(instigator, amount);
-        this.amount = amount;
+        super(amount, instigator);
+        Preconditions.checkArgument(amount >= 0, "restoration amount must be non-negative - use OnDamagedEvent instead");
```
jdrueckert, line 17: see above

src/main/java/org/terasology/module/health/events/RegisterRegenEvent.java
```diff
@@ -0,0 +1,17 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.module.health.events;
+
+public class RegisterRegenEvent {
```
jdrueckert, line 6: where is the regen registation processed? :thinking:

src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
skaldarnar: the call to `super()` needs to be the first statement in the constructor. I'd like to have the condition check first, but that's not possible.

src/main/java/org/terasology/module/health/events/RegisterRegenEvent.java
skaldarnar: nowhere yet. see commit message:

> This prepares to switch to a model and event flow better suited for the ECS and module collaboration for the regeneration feature. Both events are not used yet (neither sent nor reacted on).

src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
jdrueckert: okay, wasn't aware of that.

jdrueckert (approved):
Looks reasonable IMO. Feel free to wait for other reviewer's opinions.

PR #80, opened 2021-04-18
refactor!: improve restoration and damage event logic

skaldarnar:
## :rocket: Features and Refactorings 

- feat!: add precondition check to DoDamage and DoRestore events to ensure non-negative amounts
- refactor: decrease coupling of default restoration logic to the authority system
- refactor: perform UI updates in client system
    
Move `onMaxHealthChanged` to HealthClientSystem, and remove NUIManager dependency from HealthAuthoritySystem.    
Tested locally with 1st client hosting the game, and 2nd client connecting.

## :bug: Fixes

- fix!: get uncapped result after collecting restoration value
- fix!: get uncapped result after collection damage amount
    
The logic in place actually handles the case where the resulting value is less than 0, but due to the used getter this code path was actually dead.
Using the uncapped getter seems more reasonable here.    

## :test_tube: Tests

- test: adjust tests for damage and restoration scenarios to changed logic
- test: remove code duplication in DamageEventTest
    
As it is no longer possible to create these events with a negative base value we now assert that attempting to do this throws an `IllegalArgumentException`.    
In addition, the behavior is different if the collection phase yields a negative change value in total. In that case, the triggered event is inverted, e.g., if a restoration yields a negative restoration amount the health system sends out an `OnDamagedEvent`.

BREAKING CHANGE: Creating `DoDamageEvent` or `DoRestoreEvent` with a negative base value will now fail with an `IllegalArgumentException`.

### DIFF
--- a/src/main/java/org/terasology/module/health/events/DoDamageEvent.java
+++ b/src/main/java/org/terasology/module/health/events/DoDamageEvent.java
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.terasology.module.health.events;
 
+import com.google.common.base.Preconditions;
 import org.terasology.engine.entitySystem.entity.EntityRef;
 import org.terasology.engine.entitySystem.event.Event;
 import org.terasology.engine.entitySystem.prefab.Prefab;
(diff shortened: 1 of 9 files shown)


skaldarnar:

jdrueckert (approved):
Seems reasonable to me.

PR #81, opened 2021-04-23
feat: introduce collector events for regeneration (not yet used)

skaldarnar:
Depends on https://github.com/Terasology/Health/pull/80

I reverted 69f9ea0 and d6e709b as _registration of regeneration ids_ has to target a specific anyways, and the process is closer to the the current activation/deactivation than originally expected.

This PR mainly introduces the logic to send out and process _collector events_ for regeneration (`BeforeRegenEvent`) without actually using them yet. 

In addition, after discussion with @jdrueckert we came to the conclusion that restoration, regeneration, and damage should all use the non-negative capped results of the collector events to prevent unintentional negative effects (otherwise, it would be quite hard for a system to express something like "reduce the regeneration by 10, but not below 0"). We encourage to send out the dual event to explicitly trigger the opposite action.

image


skaldarnar:

src/main/java/org/terasology/module/health/events/BeforeRegenEvent.java
```diff
@@ -4,17 +4,31 @@
...
+ * A separate collector event is sent for each registered regeneration id. The resulting amount per regeneration id is
+ * at minimum 0 (no negative effects due to regeneration). The final regeneration amount is computed from the sum of
+ * all amounts per regeneration id.
+ *
+ * To inflict damage caused by regeneration send a separate {@link DoDamageEvent} instead.
+ */
 public class BeforeRegenEvent extends AbstractConsumableValueModifiableEvent {
 
-    private final String id;
+    final Name id;
```
jdrueckert, line 20: why is this not `private`?

src/main/java/org/terasology/module/health/events/BeforeRegenEvent.java
```diff
@@ -4,17 +4,31 @@
...
 
-    protected BeforeRegenEvent(String id, float baseValue) {
+    public BeforeRegenEvent(Name id, float baseValue) {
```
jdrueckert, line 22: why is this not `protected`?

src/main/java/org/terasology/module/health/events/BeforeRegenEvent.java
```diff
@@ -4,17 +4,31 @@
...
         super(baseValue);
         this.id = id;
     }
 
-    public String getId() {
+    public BeforeRegenEvent(String id, float baseValue) {
```
jdrueckert, line 27: see above

jdrueckert (commented):
You mention in the PR description that we discussed using the capped value. However in this PR this is not reflected yet. Is this done in a follow-up PR?

src/main/java/org/terasology/module/health/events/BeforeRegenEvent.java
skaldarnar: I don't think protected does make much sense here, as it would allow sub-classes to access it - whatever we'd need a subclass for this event for. 

The main problem here is that I sorted event into an `events` package, and systems into a `systems` package. For the systems to be able to create these events they need to be visible to them, and in this case this means the event constructor needs to be public. 

Maybe it makes sense to give another package structure a try :thinking: I'm moving everything around anyways, so let's see how this would turn out...

**tl;dr:** it needs to be public to be accessible to classes in the `systems` package.

src/main/java/org/terasology/module/health/events/BeforeRegenEvent.java
skaldarnar: Good point. I'll incorporate that in one of the follow-up PRs if that's okay with you.

src/main/java/org/terasology/module/health/events/BeforeRegenEvent.java
jdrueckert: Huh, didn't think about it that way... That might indeed be a valid insight into how we should structure packages... Looking forward to see what you'll come up with 👍

src/main/java/org/terasology/module/health/events/BeforeRegenEvent.java
jdrueckert: yes, okay

jdrueckert (approved):

PR #82, opened 2021-04-23
feat: introduce CoreHealthSystem to separate API from default features

skaldarnar:
This is just another small step towards restructuring this whole module. Expect things to move around again in later PRs.

Depends on https://github.com/Terasology/Health/pull/80 and https://github.com/Terasology/Health/pull/81.

### DIFF
--- a/src/main/java/org/terasology/module/health/core/CoreHealthSystem.java
+++ b/src/main/java/org/terasology/module/health/core/CoreHealthSystem.java
@@ -1,0 +1,35 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.module.health.core;
+
+import org.terasology.engine.entitySystem.entity.EntityRef;
+import org.terasology.engine.entitySystem.event.ReceiveEvent;
+import org.terasology.engine.entitySystem.systems.RegisterMode;
+import org.terasology.engine.entitySystem.systems.RegisterSystem;
+import org.terasology.module.health.components.HealthComponent;
+import org.terasology.module.health.events.BeforeRegenEvent;
+import org.terasology.naming.Name;
+
+/**
+ * A system adding core functionality and default behavior based on the mechanics defined in this module.
+ */
+@RegisterSystem(RegisterMode.AUTHORITY)
+public class CoreHealthSystem {
+
+    public static final Name BASE_REGEN = new Name("health:baseRegen");
+
+    /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/terasology/module/health/core/CoreHealthSystem.java
```diff
@@ -0,0 +1,35 @@
+// Copyright 2021 The Terasology Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+package org.terasology.module.health.core;
```
skaldarnar, line 4: I've put this in a separate package, although I'm not sure about it. My goal is to separate the API of Health other modules can and should integrate with from the _default logic_ served by this module. 

Other aspects that I see in this system are the short timeout before regeneration when an entity is hit. These default feature themselves use the API and serve as example implementations (at least that's my idea - "eat your own dog food" :dog: )

src/main/java/org/terasology/module/health/systems/RegenAuthoritySystem.java
```diff
@@ -111,6 +98,8 @@ private void invokeRegenOperations(long currentWorldTime) {
                 .forEach(regenEntity -> {
                     RegenComponent regen = regenEntity.getComponent(RegenComponent.class);
                     regenSortedByTime.remove(regen.soonestEndTime, regenEntity);
+                    //TODO: Add a flag to indicate the regeneration id should be removed as soon as the entity is restored to full health?
+                    //      On the other hand, a system can react to the OnFullyHealedEvent and unregister the regeneration id if desired.
```
skaldarnar, line 102: I tend toward removing this piece of logic to remove "completed" regeneration actions, but I'm not sure what the _usual use case_ would be here. I can imagine having a flag to indicate that the regen action is a "only-once" action and should be removed, but I'd like to explore how complicated the solution via `OnFullyHealedEvent` turns out.

src/main/java/org/terasology/module/health/systems/RegenAuthoritySystem.java
```diff
@@ -83,26 +83,13 @@ public void update(float delta) {
         }
     }
 
-
     /**
      * @param currentWorldTime the current in-game time in ms
      */
     private void invokeRegenOperations(long currentWorldTime) {
         // Contains all the entities with current time crossing EndTime
-        List<EntityRef> entitiesWithExpiringRegenActions = new LinkedList<>();
```
skaldarnar, line 92: This was plainly refactored using IntelliJ's "extract this" action.

skaldarnar:

jdrueckert (approved):

PR #84, opened 2021-04-27
feat!: overhaul regeneration logic

skaldarnar:
> :information_source: Please merge with **merge commit** to preserve individual commits.

This is a complete overhaul for how the **regeneration logic** works. My goal was to make this based cleanly on ECS concepts and the event flow pattern. It removes a lot of complexity, but also some optimizations that have been in place before. 

I don't expect any noticeable impact on performance right now (depends on the number of entities for which regeneration is triggered). I left `//TODO: ...` comments in place for follow ups. 

Continuing work of #82, #81, #80.

> :warning: I haven't tested each individual commit, so expect that compilation and/or logic may be broken for commits in the middle.

The biggest change is the swap of the logic of `RegenAuthoritySystem`in 6e62b9d. The commits before that lead towards the implementation of the new logic, by introducing new components or adding/changing fields on existing components. 
The commits after that are for cleanup, e.g., removing unused fields, and fixing tests.

I recommend to go through the changes [commit by commit](https://github.com/Terasology/Health/pull/84/commits).

I've tested the following: 

- [x] base regeneration of the player after taking damage (single player)
- [x] base regeneration of blocks after being damaged

BREAKING CHANGE: This changes the basics of the regeneration logic. To prevent regeneration, a system should consume the `BeforeRegenEvent` collector event. To contribute to regeneration, as system can add to the `BeforeRegenEvent` collector event. The base regeneration is controlled by the `BaseRegenComponent`. Basic regeneration needs to be explicitly registered (e.g., in a prefab or by explicit `ActivateRegenEvent` trigger event).


src/main/java/org/terasology/module/health/systems/HealthCommands.java
```diff
@@ -155,21 +157,24 @@ public String setMaxHealth(@Sender EntityRef client, @CommandParam("max") int ma
...
+        BaseRegenComponent baseRegen = clientComp.character.getComponent(BaseRegenComponent.class);
+
+        if (baseRegen != null) {
+            float oldRegenRate = baseRegen.regenRate;
+            baseRegen.regenRate = rate;
+            clientComp.character.saveComponent(baseRegen);
+            return "Health regeneration changed from " + oldRegenRate + " to " + rate;
         }
-        return "Health regeneration changed from " + oldRegenRate + " to " + rate;
+        return "FAILED no base regen component on entity " + clientComp.character;
```
jdrueckert, line 168: should we log something here or rather outside?

src/main/java/org/terasology/module/health/systems/HealthCommands.java
```diff
@@ -155,21 +157,24 @@ public String setMaxHealth(@Sender EntityRef client, @CommandParam("max") int ma
...
     @Command(shortDescription = "Show your health", requiredPermission = PermissionManager.NO_PERMISSION)
     public String showHealth(@Sender EntityRef client) {
         ClientComponent clientComp = client.getComponent(ClientComponent.class);
         HealthComponent health = clientComp.character.getComponent(HealthComponent.class);
+        BaseRegenComponent baseRegen = clientComp.character.getComponent(BaseRegenComponent.class);
         if (health != null) {
-            return "Your health:" + health.currentHealth + " max:" + health.maxHealth + " regen:" + health.regenRate;
+            return "Your health:" + health.currentHealth + " max:" + health.maxHealth + " regen:" + baseRegen.regenRate;
         }
         return "I guess you're dead?";
```
jdrueckert, line 179: should we log something here?

src/main/java/org/terasology/module/health/systems/BlockDamageAuthoritySystem.java
```diff
@@ -233,21 +237,39 @@ public void onAttackHealthlessActAsBlock(AttackEvent event, EntityRef targetEnti
...
+            if (!blockEntity.hasComponent(BaseRegenComponent.class)) {
+                BaseRegenComponent baseRegenComponent = new BaseRegenComponent();
+                baseRegenComponent.regenRate = type.getHardness() / BLOCK_REGEN_SECONDS;
+                baseRegenComponent.waitBeforeRegen = 1f;
+
+                blockEntity.addComponent(baseRegenComponent);
+            }
+
+            if (!blockEntity.hasComponent(RegenComponent.class)) {
+                //TODO: should this just send a registration event instead of creating the component on it's own?
```
jdrueckert: if we want to treat base regen like any other regen, then I'd say yes

src/main/java/org/terasology/module/health/systems/RegenAuthoritySystem.java
```diff
@@ -44,137 +29,46 @@
...
-                    entity.send(new OnFullyHealedEvent(entity));
-                }
-                entity.saveComponent(health);
-            }
+        //TODO: bring back to optimization to update each entity only once per second, but scan for
+        //      entities to regenerate every 200ms
+        regenTick += delta;
+        Instant currentTime = Instant.fromMillis(time.getGameTimeInMs());
+
+        if (regenTick > 0.2f) {
```
jdrueckert, line 55: magic number

src/main/java/org/terasology/module/health/systems/HealthCommands.java
skaldarnar: this string is printed to the game console - I'm not even sure whether that is logged automatically or not :thinking: as this is usually a user error, I don't think there's any need for extensive logging.

src/main/java/org/terasology/module/health/systems/RegenAuthoritySystem.java
skaldarnar: Yep, I'm not yet :100:% sure how to handle the "tick rate" here... But can definitely use a name constant for this :+1:

src/main/java/org/terasology/module/health/systems/BlockDamageAuthoritySystem.java
skaldarnar: yep, I'll leave this to future me when reworking the damage part or the block handling part.

jdrueckert:

deltas/engine/prefabs/player/player.prefab
```diff
@@ -8,5 +8,14 @@
     "destroyEntityOnNoHealth": true,
     "regenRate": 3,
     "waitBeforeRegen": 10
```
jdrueckert: please remove these

jdrueckert (approved):

jdrueckert:
Tests fail locally as well.
However, they also fail on develop, so I don't expect them to be related to this change, but rather generally to some access permission or loading issue with gestalt.

PR #89, opened 2021-05-16
style: fix static analysis reports (checkstyle, spotbugs, pmd)

skaldarnar:
`gradlew :modules:Health:checkstyleMain` - checkstyle integration with IntelliJ is pretty good. There's actually no need to run this manually via Gradle as IntelliJ gives the same results, and directly points to the respective locations. :+1: 

`gradlew :modules:Health:spotbugsMain` - SpotBugs with the [SpotBugs Plugin](https://plugins.jetbrains.com/plugin/14014-spotbugs) gave a mixed impression. You cannot run it directly on the module, but have to select the `src/main` source set _after_ compiling the module. However, when run via command line it shows the exception stack trace, but no link to a human readable report... nargh :roll_eyes: 

`gradlew :modules:Health:pmdMain` - Could not get this to work with IntelliJ and our configuration. When run via Gradle, I get 8 warnings which I could easily work on. IntelliJ did not find our configuration and used some default, which report 830 errors :eyes:

### DIFF
--- a/src/main/java/org/terasology/module/health/components/DamageResistComponent.java
+++ b/src/main/java/org/terasology/module/health/components/DamageResistComponent.java
@@ -3,8 +3,9 @@
 package org.terasology.module.health.components;
 
 import org.terasology.engine.entitySystem.Component;
+
 import java.util.Map;
 
 public class DamageResistComponent implements Component {
-   public Map<String,Float> damageTypes;
+    public Map<String, Float> damageTypes;
 }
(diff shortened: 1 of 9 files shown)


keturn:
> You cannot run it directly on the module, but have to select the src/main source set after compiling the module.

That lines up with what my search results were telling me yesterday. Checkstyle and PMD do static analysis on the source, Spotbugs does analysis on the bytecode.

Seems that limits Spotbugs to that sort of post-compile / pre-merge phase instead of an on-the-fly thing. Hopefully the difference in approach makes up for the inconvenience by noticing things the others wouldn't?

jdrueckert (approved):

PR #91, opened 2021-05-29
refactor: play damage sounds

skaldarnar:
Remove special handling of damage sounds for blocks, and instead make use of the default damage sound handling of the `DamageAuthoritySystem`.

The logic around playing damage sounds defined in `DamageSoundComponent` and/or `CharacterSoundComponent` was intertwined with (unnecessary?) dependencies between them. In addition, the `BlockDamageAuthoritySystem` had its own way of determining the damage sound of blocks and sending the trigger event to play it.


jdrueckert (approved):

jdrueckert:
Failing tests were failing before.

PR #92, opened 2021-06-09
feature(ecs-gestalt): Migrate Components to gestalt's Components.

DarkWeird:
Ref: https://github.com/MovingBlocks/Terasology/pull/4753

### DIFF
--- a/src/main/java/org/terasology/module/health/components/DamageResistComponent.java
+++ b/src/main/java/org/terasology/module/health/components/DamageResistComponent.java
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.terasology.module.health.components;
 
+import com.google.common.collect.Maps;
 import org.terasology.gestalt.entitysystem.component.Component;
 
 import java.util.Map;
@@ -11,6 +12,6 @@ public class DamageResistComponent implements Component<DamageResistComponent> {
 
     @Override
     public void copy(DamageResistComponent other) {
-        this.damageTypes = other.damageTypes;
+        this.damageTypes = Maps.newHashMap(other.damageTypes);
     }
 }
--- a/src/main/java/org/terasology/module/health/components/DamageResistComponent.java
+++ b/src/main/java/org/terasology/module/health/components/DamageResistComponent.java
@@ -2,10 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 package org.terasology.module.health.components;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/terasology/module/health/components/DamageResistComponent.java
```diff
@@ -2,10 +2,15 @@
...
 
 import java.util.Map;
 
-public class DamageResistComponent implements Component {
+public class DamageResistComponent implements Component<DamageResistComponent> {
     public Map<String, Float> damageTypes;
+
+    @Override
+    public void copy(DamageResistComponent other) {
+        this.damageTypes = other.damageTypes;
```
skaldarnar: Same as for the engine PR: is this a deep copy? Should this rather be something like

```java
this.damageTypes.clear();
this.damageTypes.putAll(other.damageTypes);
```

Additionally, the map might not even be initialized here :see_no_evil:

src/main/java/org/terasology/module/health/components/DamageSoundComponent.java
```diff
@@ -4,13 +4,18 @@
...
 /**
  * Contains list of damage sounds, one of which is played when entity is damaged.
  */
-public class DamageSoundComponent implements Component {
+public class DamageSoundComponent implements Component<DamageSoundComponent> {
     public List<StaticSound> sounds = Lists.newArrayList();
+
+    @Override
+    public void copy(DamageSoundComponent other) {
+        this.sounds = other.sounds;
```
skaldarnar: It looks like `StaticSound` is an immutable asset, which means that it should be fine to do a shallow copy of this list, right?

src/main/java/org/terasology/module/health/components/RegenComponent.java
```diff
@@ -57,4 +57,10 @@
      */
     @Replicate
     public Map<Name, Instant> actions = new HashMap<>();
+
+    @Override
+    public void copy(RegenComponent other) {
+        this.remainder = other.remainder;
+        this.actions = other.actions;
```
skaldarnar: Both `Name` and `Instant` are immutable, so its probably sufficient to do a shallow clone of the map only?

skaldarnar (commented):
:+1: I think this is looking good now. 

(not approving yet as we'll have to wait for the base PR anyways).

skaldarnar:
Closed by #95 (duplicate)

PR #93, opened 2021-06-24
chore: overhaul the block damage system

skaldarnar:
- doc: improve docstrings of BlockDamageRenderer
- chore: empty default implementation for RenderSystem interface
- doc: add better docstrings for BlockDamageRenderer
- feat: broadcast health changed events (was owner events)

Requires MovingBlocks/Terasology#4789

### DIFF
--- a/src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
+++ b/src/main/java/org/terasology/module/health/events/OnDamagedEvent.java
@@ -5,12 +5,12 @@ package org.terasology.module.health.events;
 import com.google.common.base.Preconditions;
 import org.terasology.engine.entitySystem.entity.EntityRef;
 import org.terasology.engine.entitySystem.prefab.Prefab;
-import org.terasology.engine.network.OwnerEvent;
+import org.terasology.engine.network.BroadcastEvent;
 
 /**
  * This event is sent after damage has been dealt to an entity.
  */
-@OwnerEvent
+@BroadcastEvent
 public class OnDamagedEvent extends OnHealthChangedEvent {
     private final Prefab damageType;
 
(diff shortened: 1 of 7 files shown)


skaldarnar:
### Open issues

- [x] `OnDamaged` is an `@OwnerEvent` and thus only visible on the authority and the client owning the entity. Therefore, no client receives the event and thus can't react on it to show any particle effects. 
  
  We may make the event a `@BroadcastEvent`, but I don't know the implications of that. There is an open card to understand networks events better.
  
  **Solved by 8c234fd**

- [ ]  On a client-only system, some default particle effects (with smoke or dust texture) are shown in addition to the expected square particles derived from the block texture. I have no idea where they are coming from :eyes: 
  **Edit:** I don't think this is related to this PR, as I've seen it also without these changes.

skaldarnar:
I'm not really sure about the change from `@OwnerEvent` to `@BroadcastEvent` I did here. Some thoughts on this:

- The server knows about the damaged entity anyways (regardless of `OwnerEvent` or `BroadcastEvent` event
- The owner is currently the only system reacting to this client-side, so `OwnerEvent` would be sufficient
- Should other clients know about whether an entity was actually damaged or not? There might be game modes that want to hide this from opponent players, but sending this event would make it easy to add a client-side modification to visualize a damage effect.
- Vice versa, not sending this event to other clients prevents them from producing additional client-side effects.

I guess, that for now it's better to leave this as `@OwnerEvent` and only broaden the scope if we have a concrete use case for this.

Edit: Reverted in 9781124

skaldarnar:
I can run the MTE tests just fine locally :thinking: 

image

jdrueckert (approved):

jdrueckert:

PR #96, opened 2021-07-11
fix: only render damage overlay on client receiving damage

### LINKED ISSUE #83: Damage Indicator shows up on Host if Client is damaged
In a multiplayer setting, when the client is damaged, the visual damage indicator is shown on the host instead of the client. This happens irrespective of whether the damage inflicted comes from the host, another client or other origins such as falling.

skaldarnar:
Fixes #83


skaldarnar:
Tests pass locally for me.

image

jdrueckert (approved):

Issue #101, opened 2023-04-22
Fix JavaDoc integration in module docs


jdrueckert:
The API sections in the module docs are currently leading to
![image](https://user-images.githubusercontent.com/29981695/233804369-cc6a1ec6-9093-4fb3-a9b1-fdd3648b05be.png)

The links in question are
- http://jenkins.terasology.io/teraorg/job/Terasology/job/Modules/job/H/job/Health/job/master/javadoc/overview-summary.html
- http://jenkins.terasology.io/teraorg/job/Terasology/job/Modules/job/H/job/Health/job/develop/javadoc/overview-summary.html

This is due to the switch to a newer jenkins (http://newerjenkins.terasology.io instead of https://jenkins.terasology.io).

The new javadocs will be available on
- http://newerjenkins.terasology.io/job/Terasology/job/Modules/job/H/job/Health/job/develop/...
- http://newerjenkins.terasology.io/job/Terasology/job/Modules/job/H/job/Health/job/master/...

once they ran for the first time which they did not do yet.

Blocks #104

Cervator:
Is this a limited issue on just one or a few modules needing it? I expect the URL will change again, maybe even during the fun a week from now. Easy enough to change a bit, especially if new URL happens later, just wouldn't want to change something per-module while on a temp setup :-)

jdrueckert:
> Is this a limited issue on just one or a few modules needing it? I expect the URL will change again, maybe even during the fun a week from now. Easy enough to change a bit, especially if new URL happens later, just wouldn't want to change something per-module while on a temp setup :-)

Fair point. I don't think we have this kind of doc structure in a lot of modules yet, so it's more of a one off fix. But if you expect it to change it soon again anyway, we can set it to blocked for now. Just have to remember to eventually unblock it again.

PR #102, opened 2023-04-22
doc: add component and ui element docs

jdrueckert:
Player Docs - UI elements and game mechanics
* Health HUD
* Damage Overlay
* Directional Damage Overlay

Modder Docs - concepts & public API
* Health
* Damage
* Regeneration
* Restoration

Developer Docs - architecture & API
* API (master) Reference
* Snapshot API (develop) Reference


jdrueckert:
Had some more discussion with @skaldarnar on module doc structure today, so I'll add some more changes before setting this back to "Ready for review".
Gist:
* **player documentation** on UI elements, effects and mechanisms introduced by activating this module
* **modder documentation** on public API (which events are processed and sent by module systems, which components can be used by module-external systems) based on common use cases
* **developer documentation** on module architecture + link to javadoc

docs/damage.md
```diff
@@ -0,0 +1,29 @@
+# Damage
+
+Damage refers to the reduction of health points caused by the affected entity itself or a third party.
+Self-inflicted causes include, for instance, falling, drowning, or poison damage.
+Damage inflicted by a third party typically involves some form of hitting the affected entity.
```
skaldarnar, line 5: I'm not sure about this separation :thinking: I wouldn't have thought of "poison damage" as self-inflicted, for example. 

Why do we think this distinction is important? 

Ideally, the _instigator_ (I guess this is what this is technically about) could carry information about the body of water the player drowned in, the exact poison or disease they are infected with, etc. The instigator might not be another player or NPC, but it could be any _entity_ in the game.

docs/damage.md
```diff
@@ -0,0 +1,29 @@
...
+
+The `DamageSystem` handles damage dealt to entities with health.
+The `BlockDamageAuthoritySystem` enables blocks to sustain some damage before getting destroyed.
+The `DamageEffectAuthoritySystem` produces block particle effect in the event of damage.
```
skaldarnar: This will be rendered as single paragraph. Should we use bullet points here?

docs/damage.md
```diff
@@ -0,0 +1,29 @@
...
+
+Send a `DoDamageEvent` to a specific entity to deal damage to it.
+
+Use the `DamageResistComponent` to configure an entity to not take damage from specific damage types.
+Examples for damage types are drowning (`engine:drowningDamage`) or explosive (`engine:explosiveDamage`) damage.
```
skaldarnar, line 14: Future: explain how to add more damage types?

docs/damage.md
```diff
@@ -0,0 +1,29 @@
...
+The `DamageAuthoritySystem` considers the settings provided by this component before an entity is damaged.
+
+Damage sounds added to an entity via the `DamageSoundComponent` are randomly selected and played when the entity is damaged.
+Sounds should be referenced as `[engine|<module>]:<soundFileName>` for sound files located in the `assets/sounds` directory of the engine or a module, for instance `engine:Slime3`.
+
+Event chain:
+* `DoDamageEvent`
+* `BeforeDamageEvent`
+* Entity damaged, health component saved
```
skaldarnar: What do you think of using italic font to indicate that this is an abstract explanation of what is happening?

```suggestion
* _Entity damaged, health component saved_
```

docs/damage.md
```diff
@@ -0,0 +1,29 @@
...
+* `OnDamagedEvent`
+
+Commands:
+* `damageResist(damagetype,percentage)`: gives resistance to damage (damagetype = all for total resistance).
```
skaldarnar, line 27: Future: Would be nice to have a command to list possible damage types.

docs/health.md
```diff
@@ -0,0 +1,8 @@
+# Health
+
+Health determines the maximum amount of damage any entity to which the health concept applies can take before, for instance, being destroyed.
```
skaldarnar:
```suggestion
Health determines the maximum amount of damage any entity can take, for instance, before being destroyed.
The health of an entity is measured in discrete steps, often called hit points (HP).
```

docs/regeneration.md
```diff
@@ -1,9 +1,10 @@
...
-Handles the natural healing of entities (and blocks).
-To activate regeneration send `ActivateRegenEvent(String id, float value, float endTime)`. 
-Health is regenerated every second. 
-Empty event `ActivateRegenEvent()` activates base regeneration of entity.
+The `RegenAuthoritySystem` handles the natural healing of entities including players, NPCs, and even blocks.
 
-To deactivate particular type of regeneration, send `DeactivateRegenEvent(String id)`. 
-Empty event `DeactivateRegenEvent()` deactivates base regeneration fo entity.
\ No newline at end of file
+To activate regeneration send `ActivateRegenEvent(String id, float value, float endTime)` for a specific entity.
```
skaldarnar, line 5: Future: Link to JavaDoc when referencing events or methods.

docs/ui-damage-overlay.md
```diff
@@ -0,0 +1,14 @@
+# Damage Overlay
+
+The `Health` module adds a visual overlay whenever the you are damaged.
+If you hurt yourself, for instance through drowning or fall damage, the default damage overlay is displayed.
+It consists of four red indicators, one on each side of your screen.
+The default damage overlay is also shown when you are hurt by an effect such as poison.
```
skaldarnar: See comment above. Here we write about "hurt by an effect such as poison" as opposed to "hurt yourself", but in the Damage section we list "poison" as example for self-inflicted damage...

skaldarnar (dismissed):
Added a few smaller remarks and minor notes, but overall looks good and improves the documentation! :+1:

docs/damage.md
Cervator: I do vaguely recall this from the past - poison is "internal" damage in a sense, unlike third party direct damage. Probably "self-inflicted" is a poor term for that. It would be nice if an internal damage effect had an _original_ trigger source that could be referenced as well.

> Player X died from poisoning, after getting too close to a Thornwood Cactus!

That might be a bit of a lofty goal for this point in time tho :-)

docs/damage.md
Cervator: I have a vague recollection that we have a module that adds a new damage type. But I have no idea what that is 🤔

docs/ui-damage-overlay.md
Cervator: This might again warrant a bit more distinctiveness (or a strive toward that at some point in the future)

* Player X died from fall damage after jumping off a cliff!
* Player X died from fall damage after being pushed off a cliff by <instigator>
* Player X died from poisoning after eating some bad puffer fish!
* Player X died from poisoning after being hit by a grave-seeker arrow shot by <instigator>!

Actually, those examples are less about hurting oneself as being about whether there is another instigator or not 🤔

docs/damage.md
jdrueckert: > Why do we think this distinction is important?

This distinction is important because the directional damage indicator depends on having a direction to indicate. If we as the player are drowning in water, suffocating due to gas, or dying to poison in our system, there is no specific direction for the damage indicator to indicate as the "instigator" is in ourselves (poison) or all around us (water/gas).
We can also make the distinction between an instigator in or all around us vs an instigator at a specific location, however I have even less of a clue how to name _that_ then...

> Probably "self-inflicted" is a poor term for that.

Yeah, naming things is not exactly a strength of mine... I'm happy about any suggestions!

> It would be nice if an internal damage effect had an original trigger source that could be referenced as well.

In terms of the directional damage indicator, I'm not sure how it would help the player to know that the cactus they originally got poisoned by, for instance is somewhere to their right...

For the player the interesting point is "why  / where from do I lose health right now?" The directional damage indicator shows them, where their attacker can be found, for instance an archer. It's important for the player to understand whether there's an attacker they might be able to defend themselves against or whether the damage is caused by an "environmental" (again maybe not the right term) effect like water, gas, poison, ...

docs/ui-damage-overlay.md
jdrueckert: see my comment above

docs/damage.md
jdrueckert: done

docs/damage.md
jdrueckert: done

docs/health.md
jdrueckert: done

docs/damage.md
jdrueckert: Created https://github.com/Terasology/Health/issues/103 to track this. Not sure we want the main docs for that in the `Health` module, though... I guess it makes more sense putting it in the engine wiki and linking to it from docs of relevant modules including `Health`.

docs/damage.md
jdrueckert: Great idea, I can write an issue for that. However, I'm wondering whether that should be a command introduced by `Health`... Damage types are an engine-level concept after all...

docs/regeneration.md
jdrueckert: Created https://github.com/Terasology/Health/issues/104 to track this. This is currently blocked by #101

docs/damage.md
skaldarnar: I like to challenge that damage types are part of the engine if the engine itself does not have a notion of, well, health....

skaldarnar (dismissed):

docs/ui-damage-overlay.md
skaldarnar: Maybe like this? We directly pick up the same wording as used in another place already.

> Self-inflicted causes include, for instance, falling, drowning, or poison damage.
> Damage inflicted by a third party typically involves some form of hitting the affected entity.

```suggestion
If you hurt yourself (self-inflicted damage), for instance through falling, drowning, or poisoning, the default damage overlay is displayed.
It consists of four red indicators, one on each side of your screen.
```

docs/damage.md
jdrueckert: fair point

skaldarnar (approved):

